PR #18, opened 2022-09-13
Kyverno policy

brandtkeller:
Logic for generation of a cluster policy from array of structs w/ execution against a live cluster


corang (approved):
👍

PR #32, opened 2022-10-17
Fix URL format for component's project page.

aj-stein-nist:
Minor and the website link is still good, but that might cause parsing issues for other tools using this example component down the line.


brandtkeller (approved):
good catch

brandtkeller:
Merging prior to a repository rename

PR #37, opened 2022-10-27
Add a Makefile

Overflow0xFFFF:
### Current behavior

Users must run `go build .` commands manually.

### Proposed changes

Add a Makefile that allows users to run `make && sudo make install` to build, test, and install lula on their systems.
This is also beneficial for future CI purposes.

The Makefile itself is heavily inspired from the one used by https://github.com/helm/helm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,68 @@
+# PREAMBLE
+#//////////////////////////////////////////////////////////////////////////////
+#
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
This sets the stage well for future development. Many thanks for the contribution.

PR #42, opened 2023-01-19
feat: add generate command

hennersz:
## Current Behavior

Kyverno policies that are generated for auditing a cluster are deleted after the audit is done

## Proposed Behavior 

This adds a generate command to output the kyverno policies from OCSAL definitions

### DIFF
--- a/src/cmd/execute/execute.go
+++ b/src/cmd/execute/execute.go
@@ -191,26 +191,26 @@ func conductExecute(componentDefinitionPaths []string, resourcePaths []string, d
 func conductGenerate(componentDefinitionPaths []string, outDirectory string) error {
 	err := os.Mkdir(outDirectory, 0755)
 	if err != nil {
-		logrus.Error(err, "error creating output directory")
+		logrus.Error("error creating output directory: ", err)
 		return err
 	}
 
 	oscalComponentDefinitions, err := oscalComponentDefinitionsFromPaths(componentDefinitionPaths)
 	if err != nil {
-		logrus.Error(err, "error getting oscal component definitions")
+		logrus.Error("error getting oscal component definitions: ", err)
 		return err
 	}
 
 	implementedReqs, err := getImplementedReqs(oscalComponentDefinitions)
 	if err != nil {
-		logrus.Error(err, "error getting implemented requirements")
+		logrus.Error("error getting implemented requirements: ", err)
 		return err
 	}
 
 	for _, implementedReq := range implementedReqs {
 		_, err := generatePolicy(implementedReq, outDirectory)
 		if err != nil {
-			logrus.Error(err, "error generating policy")
+			logrus.Error("error generating policy: ", err)
 			return err
 		}
 	}
--- a/src/cmd/execute/execute.go
+++ b/src/cmd/execute/execute.go
@@ -189,7 +189,7 @@ func conductExecute(componentDefinitionPaths []string, resourcePaths []string, d
 }
 
 func conductGenerate(componentDefinitionPaths []string, outDirectory string) error {
-	err := os.Mkdir(outDirectory, 0755)
+	err := os.MkdirAll(outDirectory, 0755)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


brandtkeller:
Thanks @hennersz for the contribution! I'll do some testing but I believe this would be a beneficial capability to have available via the CLI.

For your situational awareness - Lula is undergoing some preparatory steps for transitioning resources to full-time and establishing office-hours and other points of communication for collaboration.

src/cmd/execute/execute.go
```diff
@@ -156,6 +188,36 @@ func conductExecute(componentDefinitionPaths []string, resourcePaths []string, d
 	return nil
 }
 
+func conductGenerate(componentDefinitionPaths []string, outDirectory string) error {
+	err := os.Mkdir(outDirectory, 0755)
```
brandtkeller: If `outDirectory` already exists, this will cause an error. We likely need an initial check that evaluates the existence of the directory before attempting to create it. 

reproduce:
- Cleanup any `./out` directory ie `rm -rf ./out`
- make a directory - `mkdir out`
- run `lula generate oscal-component.yaml`

This will also fail under the circumstances where you are generating multiple policies from multiple oscal files. IF the out directory doesn't exist it will create it and generate as expected - but then each subsequent generation will fail if it attempts to make an existing directory.

hennersz: @brandtkeller I have updated this to use `os.MkdirAll` which work much like `mkdir -p` so if the directory already exists it won't error and if you supply a path with multiple directories that don't exist it will create the intermediate ones as well

brandtkeller: The above error has been resolved 👍

brandtkeller:
The only other issue that I believe would block this being a useful contribution (and wasn't part of the code added/modified here) is that the generated name for the `ClusterPolicy` is currently not valid for use with Kyverno. If you take the generated `ClusterPolicy` as it exists now and run a `kubectl apply -f <yaml file> --dry-run=server` to see the error:

```
The ClusterPolicy "42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD" is invalid: metadata.name: Invalid value: "42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

This is largely due to the name currently being uppercase from the UUID generation. We haven't encountered this yet due to it not mattering for the Kyverno Logic to execute an audit. 

I believe if we wrapped `implementedRequirement.UUID` in a `strings.ToLower(implementedRequirement.UUID)` then it looks like this will pass validation server-side. 

[Line to modify on your fork](https://github.com/hennersz/lula/blob/main/src/cmd/execute/execute.go#L275)

hennersz:
@brandtkeller I have applied the change you suggested, and installing generated policies into a real cluster now works. Sorry, probably should have checked end 2 end that this works before submitting.

brandtkeller (approved):
Changes provide minimum requirement for generation of a Kyverno `ClusterPolicy` from the OSCAL source. 

The generation can handle creation of directories that do-not exist and accommodating directories that do exist. The Policies tested were able to be applied to the cluster running Kyverno with the required CRD's without any issue.

brandtkeller:
Thank you @hennersz for the contribution. It is highly aligned with the vision of Lula and the role it plays in integrating with a policy enforcement engine.

Lula is gearing up to commit more dedicated resources into the development of the project. We will likely be evaluating current project structure and moving some logic around for clarity purposes. In the event you would like to join us on this adventure - we will ensure the issues tab is populated with where we would like to take the project and good issues for others to grab and implement. If you have questions about the addition of other logic/testing/etc - please feel free to create an issue for discussion.

PR #44, opened 2023-02-11
Kyverno v1.9.0 update

brandtkeller:
Updating the kyverno logic and dependencies from `v1.7.1` to `v1.9.0`

This is a precursor to investigations that should result in an ADR for selection of the querying capability for Lula to interact with kubernetes.


brandtkeller:
This only targets updating the existing Kyverno logic - No optimizations are included in scope.

lucasrod16 (approved):
`generate` and `execute` functionality working as expected. LGTM

PR #52, opened 2023-03-12
Migrate execute command to validate command

brandtkeller:
# description
I am making a bit of an assumption here for what I believe would be a modification for user experience. 

The `execute` command has bothered me and lends itself to some activity that might be different from an auditing activity. I also believe `validate` resonates more with the auditing adn compliance process. 

Feel free to agree or disagree. Open to closing.

# Testing
I went through the process of using the `demo/` resources to prove that validate and generate still work as intended.


src/cmd/validate/validate.go
```diff
@@ -98,29 +98,30 @@ var cluster, dryRun bool
...
-	Example: executeHelp,
+var ValidateCmd = &cobra.Command{
+	Use:     "validate",
+	Short:   "validate",
+	Example: validateHelp,
 	Run: func(cmd *cobra.Command, componentDefinitionPaths []string) {
 		// Conduct further error checking here (IE flags/arguments)
 		if len(componentDefinitionPaths) == 0 {
 			fmt.Println("Path to the local OSCAL file must be present")
+			fmt.Print(validateHelp)
```
lucasrod16, line 109: I like that you added the help/example message here for when a OSCAL file isn't passed to the command. I think that can help improve user experience

lucasrod16: It doesn't have to be in this scope of work, but it might be good to also add the help message in the same place for the `generate` command.

brandtkeller: I agree! I'm thinking after a selection on policy engine solidifies that we need issues for:
- move applicable common functions to central location for all commands (and expose publicly if valuable)
- separate generate command from the validate.go and run through initial ironing.

lucasrod16 (approved):
I like the change from `execute` to `validate`. I agree that it is a better description of the action that we are performing, and improves overall user experience. The addition of printing the help/example message for when a user doesn't provide `lula validate` a OSCAL file is also an improvement. I verified these changes with the same testing method mentioned in the description of this PR. LGTM

src/cmd/validate/validate.go
lucasrod16: I completely agree!

PR #57, opened 2023-04-24
Adding first actual ADR around OSCAL data format

brandtkeller:
Establishing Architectural Decision Record around why Lula is using OSCAL as the default data format for input and output.

Rather than implicit adoption through utilization - I believe it is valuable to explicitly define the project as an OSCAL augmenting capability. This way those who are looking to adopt OSCAL in place of other one-off formats such as excel have knowledge around what exists and the choice made by the project. 

This does not mean that Lula cannot be extended to include other data formats at a future date - but more that the default intent is towards OSCAL and moving the needle in the Risk Management and Compliance domain towards more standardized data formats. Data formats such as OSCAL where Lula is only beginning to shine light on the available automation that can be performed.


lucasrod16 (approved):
I think this captures the decision to use OSCAL well and provides important context and consequences. LGTM

Issue #59, opened 2023-04-27
Review Kyverno version constraints


brandtkeller:
Given the update to kyverno to `v1.9.0` - the generate command will generate policies that are not valid for versions of Kyverno < `v1.9.0`. 

We'll want to look at this relationship and possibly either document version compatibility or find a way to be more flexible. 

This does not impact validation - only post-generation when attempting to apply to a cluster that contains the Kyverno CRD's

brandtkeller:
Kyverno has been removed from the baseline until future integration

PR #60, opened 2023-04-27
Policy UUID relation

### LINKED ISSUE #58: Lula generate - uuid metadata
Currently we generate policies that use the `implemented-requirement` UUID as the name of the policy. It would be nice to see this data either migrated or duplicated to the resource metadata for future utilization.

brandtkeller:
DIscussion at the Policy Working Group today highlighted a need for traceability between a `ClusterPolicy` in the Cluster and the OSCAL that it may have been derived from. 

This labeling can be used by other automation later to aggregate findings for other potential collaboration. This change is mainly to drive discussion and is also low-impact. Wanted to get this through before looking at engine abstraction.

Tested application against a live-cluster to ensure Generate is creating valid ClusterPolicies

Found bug #59 in the process - not within scope to fix here.

Closes #58

### DIFF
--- a/src/cmd/validate/validate.go
+++ b/src/cmd/validate/validate.go
@@ -308,10 +308,16 @@ func generatePolicy(implementedRequirement types.ImplementedRequirement, outDir 
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name: policyName,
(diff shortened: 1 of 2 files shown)


lucasrod16 (approved):
LGTM

PR #71, opened 2023-09-27
ADR for multiple provider support and OPA validator

brandtkeller:
# Lula Update ADRs

In an effort to begin stabilizing the project (Lula), it comes as no surprise that we need to begin formalizing how the integration between Lula and providers occurs. This means re-organization of the code base to support standard structuring of native CLI logic from integrated-but-decoupled provider capabilities. 

### multiple provider support
As a baseline, we believe the project will serve the open-source community best by remaining flexible and accepting contributions towards the support of other providers as the project matures. This should be supported through clear project code structure that outlines how a provider should be accessible from the CLI and the data ingested.

There is also optionality in how data is collected (or rather the domains for which we can support collection of data) - and should a provider need data collected prior to validation (in that it cannot collect the data itself or functionality that provides static analysis), we can provide functionality for performing that collection activity.

Through investigation and learning as the project has progressed - we've seen limitations with Kyverno as an auditing capability. Particularly with cross/multi-resource inference during validation. We also want capabilities to meet the a standard for introduction into Lula. As such this ADR outlines removal of Kyverno until it can be re-integrated and contributed in a manner which is meeting of the standards of the project and evaluation of the dependencies required. 

### OPA provider
The second ADR fills the gap with regards to what functionality will be initially built and supported. As outlined in the ADR - OPA is a general purpose policy engine with support for performing validation using arbitrary structured data. This means that if the input can be structured in JSON properly that OPA can validate it using REGO. 

Primary goal is Kubernetes validation. The leap from there to cloud infrastructure is only limited by unwritten data collection logic that we believe is possible.


jeff-mccoy:
can you clarify we are referring to the GO sdk not the whole app?

Screenshot 2023-09-28 at 2 23 34 AM

brandtkeller:
> can you clarify we are referring to the GO sdk not the whole app?
> 

Good Clarification - Updated ADR to reflect use of the OPA Go SDK

brandtkeller:
Updated ADR directory to project root and added a proposed CEL ADR for future integration.

Moved Multiple Provider Support to `Accepted` barring any comments or concerns from the community.

CloudBeard (approved):
LGTM 

No questions on the proposed ADRs or changes from me. The tech changes make sense and initial validation engines are good choices to cover in and out of cluster.

PR #76, opened 2023-10-10
Oscal baseline

brandtkeller:
Tried to make this PR as small as possible.

Note: Testing structure was added but actual tests have to follow with the provider addition - otherwise it's all mocked. 

## OSCAL Baseline

The goal here was three-fold:
- Remove all existing logic for Kyverno
- Establish Baseline OSCAL library and validation logic
- Provide intended structure for providers/tests

## Important Notes

- OSCAL types were largely modified as little as possible. Top priority work in this space is to find an OSCAL-compliant location for the Lula payload. Would like to migrate to using the exported go-oscal types directly in the future. 

## Review

The diffs here are largely going to be confusing. 

I would review the files as they are suggested for update instead of looking at the diffs. The main functionality all residing in `src/cmd/validate/validate.go`


Alvin-zzz (changes requested):
Looks like a big revamp!

Alvin-zzz (approved):

src/cmd/validate/validate.go
CloudBeard, line 1: Would the long term goal to validate SSP, Assessment plan/results, and POAMS too or have those auto-update once validation on the component definition is ran/completed?

CloudBeard: Thinking in terms of

```bash
validate -filepath
```
or
```bash
validate component-definition -filepath
```

And sub out component-definition to other parts. The other side is everything is built on the component definition so if that changes everything else would need to as well. 

May be useful to error/correctness checking if we validate off SSP etc too so you know if what you currently have is right.

brandtkeller: Great question! The premise of the `validate` verb currently is only targeting a component-definition. Should we want to pivot to supporting the "validation" of other OSCAL models - we'll need to review and define what validation means in this context.

The traversal of any model is quite model specific - See `ValidateOnCompDef()` - meaning support for other models would likely include some key (IE model) to declare which model is being validated and then calling the appropriate function for traversal.

`ValidateOnSSP()`
`ValidateOnAssessmentPlan()`
etc

Still a ton of learning to do here - but I don't believe we're talking major refactor to support other models in the validation process.

CloudBeard (approved):
LGTM

PR #79, opened 2023-10-13
feat: Initial-code-scan

CloudBeard:
Setup initial codeql scan, used by zard, pepr, and uds cli.

Setup initial ossf scorecard.

Both Scorecard and vulnerabilies should be reported to the secuirty tab in GitHub.


.github/actions/golang/action.yaml
```diff
@@ -0,0 +1,10 @@
+name: setup-go
+description: "Setup Go binary and caching"
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+      with:
+        go-version: 1.21.x
```
brandtkeller: Non-critical but I would be curious if there are advantages (such as reduced cognitive burden for remembering to update the go version in multiple locations) if we substituted the `go-version` here for [go-version-file](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file)

brandtkeller (approved):
Looking forward to improving the code baseline. Thanks for this addition!

.github/actions/golang/action.yaml
CloudBeard: Makes perfect sense to me!

PR #82, opened 2023-10-15
Feat: Opa provider + Kubernetes data source + E2E test

brandtkeller:
A lot to discuss here and a lot to learn. I understand there is a lot of tribal knowledge around OSCAL and such - mainly looking for any red flags as it pertains to the code.

Core focus of review would be around:
- OPA validation logic
- Kubernetes resource collection logic

Testing made it more obvious that we may want to break down the validate logic further - more library code outside the CLI and maybe finding ways to further utilize or remove the ReportObject reference.

If I am breaking fundamental best practices anywhere - please let me know. I have a lot to learn in the space of Kubernetes API/types and more Golang best practices.

Working on the actual payload in OSCAL currently as an urgent priority - so expecting the payload location in OSCAL to change.


CloudBeard (approved):
Big fan of putting the validation rule in the oscal where you say this is how we are meeting X control. Should eliminate any confusion on proving you're doing what you just said you're doing.

E2E testing works with make test and using kind locally for me.

Only question/comment I have is using the sha instead of version tag for github actions but that can be updated when we setup rennovate to make sure its all the same versions etc.

PR #85, opened 2023-10-19
feat: initial renovate setup

CloudBeard:
Initial renovate setup. 

Using Zarf and DUBBD as starting points. 

Will need to update overtime as the project continues to expand and dependencies change.

(Currently set to every weekday, if it becomes too noisy we can lower it)


Alvin-zzz (approved):
LGTM!

brandtkeller (approved):
Looking forward to this driving maintenance for us.

PR #86, opened 2023-10-19
Codeowners update

brandtkeller:
Creating initial `CODEOWNERS` file. This will be expanded as we go but it helps establishing a baseline for branch protection.


CloudBeard (approved):
LGTM!

PR #93, opened 2023-10-21
Lula payload location

### LINKED ISSUE #15: OSCAL Schema compliance for extended oscal component definition
Update the oscal-component.yaml documents and surrounding logic to be compliant with OSCAL schema.

brandtkeller:
First iteration and closes #15 

Introduces one possible solution towards the goal of "OSCAL native" - Being that the prescribed method of the lula payload inclusion in the Component Definition OSCAL document is now using valid schema.

## optimizations
- This solution provides a mechanism for caching previous evaluations given a primary `UUID` identifier. Given that some validations may be duplicate across control groups we may gain some efficiency here.

### Notes
- `target` was added as the root of the overall payload. Language can change here but it's primarily as a workaround for trying to template yaml from a string.
  - This allows us to remove duplicate types from Lula entirely (and pull them directly from `go-oscal`.


brandtkeller:
Migrated to using upstream oscal component definition types.

Alvin-zzz (dismissed):
LGTM, we are one step closer.

CloudBeard (dismissed):
Oops wrong PR, ignore my comments.

CloudBeard (approved):
Tested locally with no issues.

PR #101, opened 2023-11-02
feat: add ability to call API as datasource for Lula validation

### LINKED ISSUE #95: API datasource
Evaluation of an API datasource for validation using available providers.

Wanting to guage interest in the ability to validate greater of body of evidence against controls satisfied for which the validation source of truth is in an API of an application.

This is useful when an application in Kubernetes does not have an entirely declarative configuration. Implementation would also hopefully be generic enough to apply to any (or most) API endpoints.

btlghrants:
Let's Dash!


Alvin-zzz (approved):
LGTM

brandtkeller (approved):
@btlghrants @TheFutonEng Thank you for your contribution to Lula. This lays a fundemental foudation for Lula to validate against known non-declarative targets/configurations that will be important for Compliance Validation.

PR #112, opened 2023-11-14
fix: append single observations to correct slice

brandtkeller:
Incorrect slice use lead to a lot of exponential file growth.


CloudBeard (approved):
LGTM

Issue #90, opened 2023-10-21
Provider evaluation of multiple objects from the datasource


brandtkeller:
Leaving a thread to continue to pull on. I believe a differentiator for Lula will be the capability to perform rich validation of many objects. 

In Kubernetes for instance, I may want to collect pods, deployments, network policies, services and write a validation that has context for all of these resources in a single decision point. 

Instead of the current design which concurrently validates all resources separately (which can still tell the same story), I believe Lula will need to be processing these as a collection.

brandtkeller:
# Resource Validation

Currently the Lula Kubernetes Domain and OPA provider operate under the current execution:
- The payload contains an array of ResourceRules
- The Kubernetes domain retrieves each of the applicable items from the cluster for each resource rule
- These items are collected into a single array of []unstructured.Unstructured
- The OPA provider converts this array into an []map[string]interface{}
  - Meaning `unstructured.Unstructured` -> `map[string]interface{}`
- The OPA provider then validates each item in the array against the rego policy


## Constraints

- This is an understood workflow - but it does not allow for the rich validation of many resources in a single validation.
  - IE In a single validation, I want to compare two resources to ensure the expected configuration matches
  - This could be done in two successive calls?
    - No, take for example the mapping of PeerAuthentications to Namespaces
      - We could have 1 -> N namespaces
      - For each namespace, we could have 0 -> N PeerAuthentications
      - if we wanted to validate that every applicable namespace has a PeerAuthentication AND every PeerAuthentication has mtls.mode set to STRICT
        - We need all namespaces
        - We need all PeerAuthentications
        - We need to ensure all namespaces have a PeerAuthentication
        - We need to ensure all PeerAuthentications have mtls.mode set to STRICT
- This is a very simple workflow - but it does not allow for the rich validation of many resources in a single validation.

## Multiple Resource Validation

- The payload contains an array of Resources
  - Each Resource contains a name and an array of ResourceRules (Adds an additional layer)
  - This allows us to collect multiple resources into groups for validation
- The Kubernetes domain retrieves each of the applicable items from the cluster for each resource rule
- These are returned as groups of items per Resource
- The OPA provider converts this array into an []map[string]interface{}
  - Meaning `[]unstructured.Unstructured` -> `map[string]interface{}`
  - Notice how this is now a collection of objects unlike previously
- The OPA provider then validates each item in the array against the rego policy

brandtkeller:
Completed in #124

PR #124, opened 2023-11-28
Resource collections

brandtkeller:
See the ADR for high-level proposal.

Introducing the ability to collect many resources into named groups that can then be used in Rego for cross-resource validation. 

This was a constraint we found when wanting to do rich validation given a live-environment and wanting to do comparisons against many resources. Global cluster context is a relevant example - other examples may be API calls which require mapping outputs together to get a full-system view. 

Added a `success` E2E test - may consider a failure scenario if it is deemed useful. Feels largely covered by singular fail scenarios. 

Add questions/concerns/critiques. The team will find a time to mob review and ensure everyone is on the same page.


mike-winberry (dismissed):

CloudBeard (dismissed):
Big fan of moving to multi-resource for rich validation. I think this will become more useful as we start to incorporate more resources.

Also tested locally and didn't have any issues. Looks good to me!

brandtkeller:
Closes #90

mike-winberry (approved):

PR #128, opened 2023-11-30
Lula tools command

brandtkeller:
Adds a `lula tools` command as a foundation for future OSCAL helper commands.

Implements a `uuidgen` command for random and deterministic uuid generation.

Replaces current uuid logic with go-oscal wrapper.


Alvin-zzz (approved):
LGTM!

PR #129, opened 2023-11-30
Fix: Jeff probably doesn't need to be required on every PR

brandtkeller:
Pretty insubstantial either way but just removing Jeff from EVERY PR.


jeff-mccoy (approved):
lgtm

Issue #139, opened 2023-12-01
no such file or directory: ./lula


chadmcrowell:
In the getting started guide, running the `./lula validate ./demo/oscal-component.yaml` command from within the `lula` directory resulted in the following error:

```bash
$ ./lula validate ./demo/oscal-component.yaml                                                                                                                    [12:37:10]
zsh: no such file or directory: ./lula
FAIL
$ pwd                                                                                                                                                           [12:41:24]
/Users/user/lula
```

please advise.

brandtkeller:
Great call out @chadmcrowell ! This was also recently mentioned in #130 - we need to update the instructions. Barring any other movement I will get this fixed asap!

Intent: 

After the `make build` step the binary is more accuretely located in the created `bin/` directory within the Lula project. 

So `./bin/lula validate ./demo/oscal-component.yaml` would be the correct execution if all other dependencies have been met.

brandtkeller:
Updated the README instructions for the demo. Please feel free to re-open this issue if any problems persist.

PR #140, opened 2023-12-01
fix: should fix getting started command in readme

CloudBeard:
Updated README to the correct command to run lula validate.

Should solve https://github.com/defenseunicorns/lula/issues/139


brandtkeller (approved):
Solid! Going this with a workflow modification before closing out the associated issues.

Issue #130, opened 2023-11-30
chore: Update Readme Demo Steps


TristanHoladay:
## Description

The current readme has a list of steps for getting started with Lula. Some of these steps and their output no longer match what is actually occurring when they are applied.

```
tristan@tristan-NUC13ANHi7:~/repos/lula$ kubectl apply -f ./demo/namespace.yaml
namespace/validation-test created

tristan@tristan-NUC13ANHi7:~/repos/lula$ kubectl apply -f ./demo/pod.fail.yaml
pod/demo-pod-fail created

tristan@tristan-NUC13ANHi7:~/repos/lula$ ./lula validate ./demo/oscal-component.yaml
bash: ./lula: No such file or directory

tristan@tristan-NUC13ANHi7:~/repos/lula$ ./bin/lula validate ./demo/oscal-component.yaml
OPA provider validating...
UUID: 42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD
	Status: not-satisfied
	
tristan@tristan-NUC13ANHi7:~/repos/lula$ kubectl apply -f ./demo/pod.pass.yaml
pod/demo-pod created

tristan@tristan-NUC13ANHi7:~/repos/lula$ ./bin/lula validate ./demo/oscal-component.yaml
OPA provider validating...
UUID: 42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD
	Status: not-satisfied
```
## ToDos
- [x] update steps with  `./lula` to `./bin/lula`
- [x] show new output for pod failing
      OPA provider validating...
      UUID: 42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD
	Status: not-satisfied
  
- [ ] add step to delete the failed pod before running validate for the passing pod
   `kubectl delete -f demo/pod.fail.yaml`

brandtkeller:
Updated in #142 

Changed the behavior of the demo such that the pod is updated in-place and thus a `kubectl delete` is not required.

PR #142, opened 2023-12-01
Fix: demo workflow and README content

### LINKED ISSUE #139: no such file or directory: ./lula

brandtkeller:
Closes #139 #130

Fix the demo to update the pod for validation to be updated in-place. This will prevent some confusion about multiple pods (IE if I have both a compliant pod and non-compliant pod - this still fails properly). 

Add information to the demo about the assessment results file.

Remove API as a future consideration as we have initial support for that already to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,11 @@ In the standard CLI workflow:
 #### Dependencies
 
 - A running Kubernetes cluster
+    - Kind
+        - `kind create cluster -n lula-test`
+    - K3d
+        - `k3d cluster create lula-test`
+- kubectl
 - GoLang version 1.21.x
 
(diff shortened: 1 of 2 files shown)


CloudBeard (approved):
Demo changes look good!

PR #151, opened 2023-12-06
deps: update go-oscal to use the released tag

brandtkeller:
Update to use the tag releases of `go-oscal`


CloudBeard (approved):
Updated to go-oscal v0.0.1 - looks good!

PR #160, opened 2023-12-08
Feat: add evaluation for assessment results files

brandtkeller:
Add evaluation command to lula for comparing the findings of two results. Opening in draft to keep ideas flowing early.

PR #163, opened 2023-12-08
Feat: Aggregate by control-id

### LINKED ISSUE #161: Aggregate findings by control-id
Given the information and understanding about the assesment results model - we will want to aggregate the findings (and the result of said finding) by control-id/target-id

brandtkeller:
Closes #161 

Removing temporary objects between component-definitions and the creation of assessment results. Now we generate the required findings and observations during validation directly and pass them to generate the assessment-results model around. 

This also unblocks #160 as it makes `control-id` unique - meaning we can compare two results and expect the state of a single finding to reflect all validations against that control-id.


CloudBeard (approved):
Changes look good! Also cool to see the UUID code from go-oscal make it into Lula.

PR #171, opened 2023-12-11
feat: Add Initial release process

CloudBeard:
Added initial release process and doc. Follows same workflow as go-oscal.

Currently includes config for Homebrew but waiting to see if we may need to drop before initial release. Switching to GitHub Apps from PAT.


.goreleaser.yaml
```diff
@@ -0,0 +1,71 @@
+before:
+  hooks:
+    - go mod tidy
+
+# Build a universal macOS binary
+universal_binaries:
+  - replace: false
+
+# Build the different combination of goos/arch binaries
+builds:
```
brandtkeller, line 10: Let's add in our `ldflags` here so as to get it automatically included in our binaries

```
ldflags:
      - -s -w -X github.com/defenseunicorns/lula/src/config.CLIVersion={{.Tag}}
```

brandtkeller:
Proactively established the tag rule for `*` on this repository.

CloudBeard:
Removed Homebrew setup. To use GitHub Apps for permissions we have to get a token from the app ID and app secret. That token dosen't have permissions to upload the artifacts to GitHub.

May be possible to break that apart with Goreleaser pro and templates.

brandtkeller (approved):
Look solid - Looking forward to getting a release out here in the near future.

PR #178, opened 2023-12-15
fix: 47 add automated tests (partially)

CloudBeard:
Add some unit tests for the uuid command and tools command.

Did not add for for lint. Need to wait for a new release of go-oscal and rework it after we bump the dependency. Issues with the way it handles errors currently so it breaks tests.

Didn't add for the lula validate command since the oscal version may/will change. 

Will add those in a separate PR.

Created a lulatest package to house some common variables and functions that may make testing long term more uniform. Copied a lot of how go-oscal is setup too to keep tools uniform.

Did not update the makefile to include the unit tests as the makefile in main will grab all ready.


src/cmd/tools/uuid_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+    "github.com/defenseunicorns/lula/src/test/lulatest"
+    "github.com/defenseunicorns/go-oscal/src/pkg/uuid"
+    "github.com/spf13/cobra"
+
+)
+
+func toolsCmdFactory() *cobra.Command {
+    // Create a new instance of cobra.Command
+    var newCmd = &cobra.Command{
```
brandtkeller: Love the intent - but we're establishing two sources of truth for a single command (This code here duplicates the actual command code.

If we move some logic around in the actual command we can essentially import it here - given this test file exists in proximity to the command - it wouldn't have to even be public.

src/cmd/tools/uuid_test.go
```diff
@@ -0,0 +1,78 @@
...
+        },
+        {
+            name: "Test the uuidCmd with too many arguments.",
+            args: []string{"https://lula.dev", "https://lula2.dev"},
+            wantErr: true,
+        },
+
+    }
+// loops through the tests and tests checks if the wantErr is true/false
+for _, tt := range tests {
```
brandtkeller: no change required - I see you found the inner-workings of table driven testing and the use of `t.Parallel` - nice work.

brandtkeller (changes requested):

src/cmd/tools/uuid_test.go
brandtkeller: For those not-tracking and wanting to see the specifics:

- https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721
- https://go.dev/wiki/LoopvarExperiment

CloudBeard:
Changing back to draft while working through errors.

src/test/intergration/lint_intergration_test.go
```diff
@@ -0,0 +1,83 @@
...
+			wantOutput: "",
+		},
+	}
+	
+	// loops through the tests and tests checks if the wantErr is true/false
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			cmd := exec.Command("../../../bin/lula", tt.args...)
```
brandtkeller: For discussion - this establishes a dependency that we build the lula binary for use in testing. Not inherently against that but would like to research options.

CloudBeard: I think gomock/testify can work in this situation to avoid that build dependency. Ill test and put something up once its working.

CloudBeard:
Tried gomock/testify and could only do a mock situation, basically create how it should work/fail. If the commands change or their functionality then the mock would still pass but the command may not work the same anymore.

Closest to live I could get was to create a geter function for root cmd so it could be used when the cmd package is imported but without making the root command itself public. 

Created a fatal wrapper function since the main issue I was running into is the app exits on message.Fatalf this way on the commands that we don't have direct access to their functions (lint, uuidgen, mostly commands straight from go-oscal but maybe some from the shared zarf libraries in the future) we can pass that to the fatal wrapper which checks if it is in test mode, if so use the LogWriter function already in message.go and add in the test Test Mode Error Encountered and the error (easier to look for the error in tests) or if test mode is false it continues as normal and just exits.

Did run into concurrency issues running the tests/subtests in parallel while using the message.LogWriter. Open to suggestions there, main goal was not to add too much complexity for a couple of test cases but could be worth investigating the logging portion more.

CloudBeard:
Closing issue as https://github.com/defenseunicorns/lula/issues/450 will encompass the testing for this and more.

PR #209, opened 2024-01-12
Fix: assessment result file provenance

### LINKED ISSUE #206: Assessment results file produced does not contain top-level key
> Failed to create validator: expected model to have 1 key, got 4

The generated assessment results file contains the contents of the model but not the top-level key. Does not pass linting.

brandtkeller:
Closes #206 

Assessment results file would not pass linting as it did not contain the top level "assessment-results" key. This adds the fix and extends a test to include checking for proper append to file AND linting the output to ensure our reports are valid OSCAL.


CloudBeard (approved):
Awesome catch and fix!

PR #230, opened 2024-01-24
fix: Updating readme and oscal file to reflect a better demo

brandtkeller:
- Removes the wait from the demo oscal - new bug found that can cause hiccups
- Update the OSCAL to be more descriptive of what is expected
- Update the README.md to include expected output and fix the validate command flag missing


CloudBeard (approved):
Changes look good. Big fan of changing it to Lula - The Compliance Validator to match the repo description.

PR #253, opened 2024-02-09
Support for named resources

### LINKED ISSUE #249: ResourceRule named resource option
## Feature

Add a `name` field to the `ResourceRule` object such that individual resources can be referenced. This allows for a more finely scoped validation to occur when we potentially only need to evaluate one specific resource. 

## Alternatives
To-date we can accomplish this already in Rego with the `some` function and identifying information.

brandtkeller:
In order to reduce the processing payload when opa begins validation - when a specific kubernetes resource is known by name we can support filtering by any matching names.

Currently this requires strict matching - will add issue to backlog to consider wildcards for names. 

Closes #249


src/pkg/common/kubernetes/resource.go
```diff
@@ -59,38 +40,77 @@ func QueryCluster(ctx context.Context, resources []types.Resource) (map[string]i
...
-		Group:    group,
-		Version:  version,
-		Resource: resource,
+		Group:    resource.Group,
+		Version:  resource.Version,
+		Resource: resource.Resource,
 	}
+	collection := make([]map[string]interface{}, 0)
+	// Query all namespaces in one execution
+	if len(resource.Namespaces) == 0 {
```
meganwolf0: Not that this is super important, but instead of essentially repeating lines 63-81 down in 85-104, you could have something like 
```go
namespaces := []string{""}
if len(resource.Namespaces) != 0 {
  namespaces = resource.Namespaces
}
for _, namespace := range namespaces {
...
}
```

brandtkeller: Agreed! Updated the issue to reflect a new direction I believe is more provenant. Planning to rewrite and have available tomorrow.

src/pkg/common/kubernetes/resource.go
```diff
@@ -59,38 +41,63 @@ func QueryCluster(ctx context.Context, resources []types.Resource) (map[string]i
...
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, err
+		}
+
+		// Reduce if named resource
+		if resource.Name != "" {
+			// requires single specified namespace
+			if len(resource.Namespaces) == 1 {
```
meganwolf0, line 79: Is there a reason to limit this to where namespaces are specified? I know some resources will have shared names across different namespaces (for instance the `private-registry` secret across all bb namespaces) - and I think the  `GetName()` supports all namespaces - just thinking about future flexibility here, not that I can currently think of a use case for grabbing all those secrets

meganwolf0:
Reviewed the code and docs, tested with some tempo validations using the named resource, and it works great! The thing that tripped me up was the named resource returning as a single object instead of an array with only one object. I'm kind of torn on consistency here, it's weird to get an array with a single element but also a bit of a context switch to have both single objects and arrays returned as resources. I think my comment questioning the possibility of having multiple named resources (e.g., across different namespaces) is still open and valid to the dataset type being returned, but there's no real use case for this at the moment so might be fair to bump this to a backlog thing.

brandtkeller:
> Reviewed the code and docs, tested with some tempo validations using the named resource, and it works great! The thing that tripped me up was the named resource returning as a single object instead of an array with only one object. I'm kind of torn on consistency here, it's weird to get an array with a single element but also a bit of a context switch to have both single objects and arrays returned as resources. I think my comment questioning the possibility of having multiple named resources (e.g., across different namespaces) is still open and valid to the dataset type being returned, but there's no real use case for this at the moment so might be fair to bump this to a backlog thing. 

I believe it's worth considering and performing more discovery to evaluate value. I went back and forth and should probably update the description of the PR to capture that I aligned to 'kubectl' for the purposes of the returned payload. IE 'kubectl get po -A' is a list vs 'kubectl get po podname -n namespace' is a single object.

meganwolf0 (approved):
looks good to me

src/pkg/common/kubernetes/resource.go
meganwolf0: resolved as alignment to kubectl and potential further evaluation in the future

CloudBeard (approved):
Changes look good to me!

PR #263, opened 2024-02-16
Testing reusable modules and repeatable Rego generation

meganwolf0:
The intent of this PR was to look at trying to DRY out the rego code a little more, add some reusable components, and just keep an eye toward making the lives easier of people that are going to find themselves writing this stuff. There's a couple things contained:
1) `lula tools regogen` creates a rego template, optionally passing it a resource kind will template with that structure (e.g., `lula tools regogen Deployment` - The hope was to create a kind of framework that ensures outputs are common and certain reusable functionality (like exemptions) can be re-used. This skeleton also hopefully indicates a few places for a user to change to meet logic ("** Change this **"). Obviously we have a really small sample set right now so this template may not cover all cases and perhaps there's different flavor templates for different types of controls.
2) Shared modules in the code - as an addendum to the above statement, the reusable scripts have to actually be accessible. I thought about putting it in our codebase, but it seems like transparency is key so ended up adding a type of backmatter to hold those modules. This ended up touching quite a bit of code but really just a proof of concept for what it could look like. This part was tested out with `Istio-common.yaml` which is on the `istio-common` branch of compliance-artifacts.

### DIFF
--- a/src/cmd/tools/rego.go
+++ b/src/cmd/tools/rego.go
@@ -1,0 +1,88 @@
+package tools
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/defenseunicorns/lula/src/config"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+	"github.com/spf13/cobra"
+)
+
+var regoHelp = `
+To create a new template for Rego Validation:
+	lula tools regogen
+
+To create a template given some resource kind:
+	lula tools regogen Deployment
+`
+
+var regoTemplate = `
+package validate
+import data.common
+
+result := get_allowances(input.{{ . | lower }}s)
+
+get_allowances(resources) = result {
+  valid := [name | resource := resources[_]; allowed(resource); name := common.print_resource(resource)]
+  invalid := [name | resource := resources[_]; not allowed(resource); name := common.print_resource(resource)]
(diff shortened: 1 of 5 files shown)


brandtkeller:
First pass - I certainly believe there is value in the shared modules. 

Implementation looks great - we'll likely want to think more about how we handle some of these commons functions - Known use-cases such as how we can combine modules across component files as we aggregate into a larger assessment-plan, how we handle for more domains than Kubernetes, and what implementation looks like without shared modules. 

What I really like is exploring more of the rego extensibility that you showed (maybe as a byproduct) for capturing `Passing` and `Failed`.

We need a concrete structure (which has been `validate` to ultimately decide `satisfied` vs `not-satisfied` but honestly if we kept the `resultMap` as `map[string]interface{}` then we could port over a lot of generic capability that we can expose in the `assessment-result` as evidence - enabling people to create their own evidence collection objects. This feels like a great immediate add vs other functionality that might require some more refinement - let me know if you need me to explain further as it is probably a pretty small feature set that wouldn't need any additional testing or functionality.

brandtkeller:
#265 created to capture the transient data vision for generic information that could be extended from Rego key/value results.

meganwolf0:
Since this was a prototype and much of the code has changed, this PR is really no longer applicable. If a future use case supports adding re-usable Rego functions this may be good to reference.

PR #267, opened 2024-02-23
Initial payload output structure definition

### LINKED ISSUE #265: Evidence mapping from OPA provider to Assessment-Results
Given the current implementation of Lula - we are no longer iterating over single resources to perform validation. Rather Lula wholistically collects all items to be assessed and performs validation of a payload of resources (1 -> n) using a Rego policy. 

This means that the current representation of items passing and failing do not serve a purpose. We should remove these from the reports.

Given Rego that establishes other key/value pairs in the [resultMap](https://github.com/defenseunicorns/lula/pull/263/files#diff-be4eb6e7e0cc9366b2eb4426ef1e44f26824b58ab104d9c8f33b775656f61092R173), we should:
- Require the `validate: bool` as a strict requirement and document it as such
- Allow for iterating over all other keys generically and mapping that data to a string that can be represented in an `observation` object in the `assessment-results` report. A generic map allows for the user to extend data available in the `resultMap` to include other evidence they might find valuable. Documenting how to accomplish this would be a great benefit to the OPA provider.

meganwolf0:
Added some initial structure for pulling more data from OPA queries, including specifying the actual boolean "validate" field to be evaluated, as well as additional observations that can be output in "remarks" section of observations. Also added a test for the field. 

Current limitation is that the field from rego only support string data - this is really more of a go constraint but I'm sure if we wanted to support additional types you could expand the definition of an "output.observation".


src/pkg/providers/opa/opa.go
```diff
@@ -121,39 +121,53 @@ func GetValidatedAssets(ctx context.Context, regoPolicy string, dataset map[stri
...
-	regoCalc := rego.New(
-		rego.Query("data.validate"),
+	// Get validation decision
+	validation := "validate.validate"
+	if output.Validation != "" {
+		validation = output.Validation
+	}
+
+	regoCalcValid := rego.New(
+		rego.Query(fmt.Sprintf("data.%s", validation)),
```
brandtkeller, line 133: I'd like to weigh the pro/cons for splitting out this logic into multiple rego evaluations (as you have here) vs potentially a single evaluation that results in `resultValid` containing  a larger slice of items that we can iterate over. 

IE as it stands now - with the scenario you added we see `resultValid` contain `[{[true] map[]}]` for `data.validate.validate` which is appropriate. Then to my understanding we execute another `rego.eval()` for each obversation, which in the validation that has output results in `[{[hello world]map[]}]`.

What am I curious about is if we take it up one level on the initial `rego.eval` to the package and use for example `data.validate` we get `resultValid` containing `[{[map[test:hello world validate:true]] map[]}]`. If we know what the `validate` target is (either default or defined) - then could we not just iterate over all remaining elements in the slice and use those as `observations` and remove the need to re-evaluate the entire dataset for each?

meganwolf0: Yeah, I did think about/try to do this as a single evaluation then pull the items needed. The issue is when those fields are in indeterminate json structure, so there's really not a clean way for go to handle that (thinking especially when validate isn't a first level field, e.g., validate.result.validate or something). You have to like marshal and unmarshal each structure for it to become usable data - then at the end of the day you still need to know what the types you are expecting so it's just pretty ugly. (I can throw up a competing PR here to delve down this path, but when I was working this yesterday that was my first impulse and after writing some lines I ran away pretty fast, ha!)

If we decided to pre-determine structure as a requirement, then agreed a single eval would probably be more appropriate - I was just thinking we maybe would like to be flexible there. But obviously the world is our oyster as far as design choices go!

If you're concerned about the time it takes to run an `eval` that's certainly something we could try and understand more. I kinda wasn't clear on latency requirements for this.

brandtkeller: You raise a good point that this reduces the surface of marshalling/unmarshalling and a lot of handling of indeterminate data. That coupled with this being an optional field that people can opt-in to makes the surface of required optimization low at this point in my opinion.

src/types/types.go
```diff
@@ -24,6 +25,18 @@ type Payload struct {
...
+	Output    Output     `json:"output" yaml:"output"`
+}
+
+type PayloadAPI struct {
+	Requests []Request `mapstructure:"requests" json:"requests" yaml:"requests"`
+	Rego     string    `json:"rego" yaml:"rego"`
+	Output   Output    `json:"output" yaml:"output"`
+}
+
+type Output struct {
```
brandtkeller, line 37: Documentation can go a long way here - but if we were to step back and think about this from another providers context (that doesn't exist today) - are the names of these fields intuitive and flexible for use? 

words are hard...

src/pkg/providers/opa/opa.go
```diff
@@ -121,39 +121,53 @@ func GetValidatedAssets(ctx context.Context, regoPolicy string, dataset map[stri
...
-				matchResult.Passing += 1
-			} else {
-				matchResult.Failing += 1
-			}
+		resultObv, err := regoCalcObv.Eval(ctx)
+		if err != nil {
+			return matchResult, fmt.Errorf("failed to evaluate rego policy: %w", err)
+		}
+		// To do: check if resultObv is empty - basically some extra error handling if a user defines an output but it's not coming out of the rego
+		if matched, ok := resultObv[0].Expressions[0].Value.(string); ok {
```
brandtkeller: Given this is supporting data for an observation - might be a good time to add debug level logging to aid in the debugging process. If something is wrong in the execution of a single item there isn't anything in the output to indicate a problem occurred.

src/types/types.go
meganwolf0: Yeah, I'm not sure if the output here would be reusable across other providers - mainly just because I don't actually know what the output from other providers would be. It seemed like `Payload` was pretty particular to OPA though (just thinking the `Rego` field specifically), so wondering if we can punt the thought exercise til we have another provider to consider? Although of course if you have particular words that you think would be more extensible/resonate better I am certainly not attached to these words!

meganwolf0:
@brandtkeller - I added some debugging to the validation and observation OPA results, just in case there's some unexpected values there, the whole thing shouldn't crash out, it will just result in a "not-satisfied" evaluation (edit: it's only "not-satisfied" if validate isn't recognized as boolean, which I think was the behavior before, debugging now just tells you why). Also, updated test case for output. How fine-grained are we making these tests? This one really just checks that the `outputs` section with various parameters doesn't fail, but no checks on if the correct things are output. I've just been checking that by examination in assessment-reports...

meganwolf0:
Updated the docs as part of this PR, moving to ready for review

docs/opa-provider.md
```diff
@@ -26,6 +26,31 @@ Example:
         }
 ```
 
+Optionally, an `output` can be specified in the `payload`. Currently, the default validation allowance/denial is given by `validate.validate`, which is really of the structure `<package-name>.<json-path-to-boolean-variable>`. If you have a desired alternative validation boolen variable, as well as additional observations to include, an output can be added to payload such as:
```
brandtkeller:
```suggestion
Optionally, an `output` can be specified in the `payload`. Currently, the default validation allowance/denial is given by `validate.validate`, which is really of the structure `<package-name>.<json-path-to-boolean-variable>`. If you have a desired alternative validation boolean variable, as well as additional observations to include, an output can be added to payload such as:
```

docs/opa-provider.md
```diff
@@ -26,6 +26,31 @@ Example:
...
+```yaml
+ - provider: "opa"
+    domain: "kubernetes"
+    payload:
```
brandtkeller: I updated documentation in #253 - but wanted to mention that this is an outdated payload that pre-dated `[]Resources` - forgot to update docs then - oops

docs/opa-provider.md
meganwolf0: updated typo

docs/opa-provider.md
meganwolf0: updated this documentation example with the correct structure

meganwolf0:
@brandtkeller - Merged main and from my perspective this is ready to go, might be nice to get other eyes on though if you think other thoughts would be good to evaluate the implementation. Also, there's one convo I'll leave to you to resolve if you feel it is.

src/test/e2e/outputs_test.go
```diff
@@ -0,0 +1,68 @@
...
+			if err = config.Client().Resources().Create(ctx, pod); err != nil {
+				t.Fatal(err)
+			}
+			err = wait.For(conditions.New(config.Client().Resources()).PodConditionMatch(pod, corev1.PodReady, corev1.ConditionTrue), wait.WithTimeout(time.Minute*5))
+			if err != nil {
+				t.Fatal(err)
+			}
+			return context.WithValue(ctx, "test-pod-outputs", pod)
+		}).
+		Assess("Validate Outputs", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
```
brandtkeller, line 35: I'd like to see more specificity in the body of this test. `validate` is not a pass/fail from the command line exit code status - so additional scrutiny in testing to target a finding state or even a specific remark given the new functionality would suffice.

src/test/e2e/outputs_test.go
```diff
@@ -0,0 +1,68 @@
...
+				t.Fatal(err)
+			}
+
+			// Write report(s) to file to examine remarks
+			report, err := oscal.GenerateAssessmentResults(findingMap, observations)
+			if err != nil {
+				t.Fatal("Failed generation of Assessment Results object with: ", err)
+			}
+
+			err = validate.WriteReport(report, "assessment-results-outputs.yaml")
```
brandtkeller: This is going to create a file that is git-ignored. For sake of cleanliness and local-dev maybe we should standardize on cleaning up these files in the `teardown` as done in `pod_validation_test.go`.

brandtkeller (changes requested):

src/test/e2e/outputs_test.go
meganwolf0: Thanks for the feedback - I definitely struggled to write good tests here. Really, I wanted to check that the validation responded properly and the observations got printed out. So I modified a bit to just check those data objects since really it's not a matter of checking that the report gets generated. It is a little hokey because it doesn't seem like there's an actual link from the validations UUIDs to observations where remarks are. Please look at the changes and let me know if there's a better structure to try and meet the intent.

src/test/e2e/outputs_test.go
meganwolf0: Removed this section as I don't think we need to check the `WriteReport` function, but can add back in if you think otherwise

src/test/e2e/outputs_test.go
brandtkeller: Appreciate the updates - more tangible as far as validating the execution worked as expected.

brandtkeller (approved):

PR #270, opened 2024-02-23
Generate Command Structure

brandtkeller:
Creation of a `generate` command skeleton and thoughts for how to generate each of the required artifact templates. 

Intent would be that each sub-command under generate could be run imperatively as well as from a manifest (more dissection to do here). 

Idea here being that there will be a lot of nuance for imperative runtime flags and having sub-commands will allow for more-easily breaking out flags as required - as such I put a structure in place for accomplishing this.

Not yet documented:

A hybrid workflow of human-in-the-loop and automation is essential to success here. We need to create workflows that accounts for an artifact existing and performs some action when the artifact already exists to retain or update current data.


CloudBeard:
Just thought about this, what about adding in flags for compliance baselines for component-definition where you can pick NIST 800-53 High, Mod, Low or FedRAMP High, Moderate, Low (Each have profiles) and then add in IL4/5/6 if wanted.

The IL piece may need to be listed out in a manifest since there isn't a "profile" specifically for those.

brandtkeller:
Closing as completed in another PR

PR #295, opened 2024-03-06
feat(cmd): #281 Create upgrade command implementing go-oscal revise. 

### LINKED ISSUE #281: Implement go-oscal revision command into a tools upgrade command
## Intent
Lula can identify when a new version of OSCAL is supported and provide a command for upgrading an existing OSCAL artifact. 

## Dependencies
https://github.com/defenseunicorns/go-oscal/issues/147

## Objectives
- Implement go-oscal revise as a helper command under `lula tools` for performing the upgrade.
- Given support for the issues dependency above - log/inform users/systems when the oscal artifact being validated could be upgraded
- Add information for how to perform the upgrade.

mike-winberry:
- feat: `lula tools upgrade` command
  - version defaults to latest supported by go-oscal
  - output file defaults to in place upgrade of input file
  - takes optional validation result file path to output a result
- feat: `uprade` command will warn if not upgrading to latest version.
- feat: `lint` command will warn if not the latest version.
- feat: `lint` command now has optional `-r` `--result-file` param that will output the `ValidationResult` if defined.
- chore: update `lula tools lint` with changes to ValidationCommand
- chore: upgrade go-oscal to v0.2.5
- docs: add disclaimer for visual changes to yaml while maintaining object equality when using the upgrade tool.
- fix: `scenarios/pod-label/component-definition.yaml` to pass `lula lint`.


meganwolf0:
So I tried this out on the tempo validation I've been working and it had the mildly annoying result of moving all the fields around and smashing down my formatted `description` field with all the payload data down into a single line... Wondering if that could be pretty printed out and ideally the backmatter could still be at end of the oscal file? I'm not entirely sure if the intended use of this is to create a file a user would continue to work with, so also maybe some documentation to go along with the intent/use case specified?

brandtkeller:
> So I tried this out on the tempo validation I've been working and it had the mildly annoying result of moving all the fields around and smashing down my formatted `description` field with all the payload data down into a single line... Wondering if that could be pretty printed out and ideally the backmatter could still be at end of the oscal file? I'm not entirely sure if the intended use of this is to create a file a user would continue to work with, so also maybe some documentation to go along with the intent/use case specified?

hhmm... yeah this makes a lot of sense that it would do this given the underlying sting datatype and how we write validations. Unfortunate side-effect - if there is anything we can do to retain that structure then great - otherwise it certainly adds more evidence to needing a system that is more compatible.

mike-winberry:
> So I tried this out on the tempo validation I've been working and it had the mildly annoying result of moving all the fields around and smashing down my formatted `description` field with all the payload data down into a single line... Wondering if that could be pretty printed out and ideally the backmatter could still be at end of the oscal file? I'm not entirely sure if the intended use of this is to create a file a user would continue to work with, so also maybe some documentation to go along with the intent/use case specified?

I will create a story to look into it and see what I can do today. My understanding right now as to why, is that the golang map is unordered, and since json and yaml are considered to also be unordered data, the marshalling and unmarshaling process doesn't guarantee order. I have some ideas on how to deal with that but will have to play a bit in go-oscal today and see what works out. 

I think another thing that could be done is looking at yaml formatters and run one before output. (for the string issue).

src/test/e2e/pod_validation_test.go
```diff
@@ -91,11 +94,22 @@ func TestPodLabelValidation(t *testing.T) {
 				t.Fatal("Failed to append results to existing report")
 			}
 
-			validator, err := validator.ValidateCommand("sar-test.yaml")
+			revisionOptions := revision.RevisionOptions{
```
brandtkeller: Captures intent of the artifact version upgrade. Would like to offer an alternative this workflow:

Could we instead upgrade the component-definition defined at the beginning of the `Assess()` function? (Are there any constraints here?)

My thinking is that this would become a incredibly valuable E2E test if we know that writing and executing validations does not get corrupted by the `lula tools upgrade` process as we continue to develop. Our test suite to-date has confirmed that the existing structure works - so adding this should mean we know if any modification to validation or upgrade breaks any part of the E2E lifecycle. 

Open to feedback.

brandtkeller (changes requested):

src/test/e2e/pod_validation_test.go
mike-winberry: That component definition is currently failing lint so I am going to see if I can fix it. 
```yaml
valid: false
timeStamp: 2024-03-12T09:29:15.379401593-07:00
errors:
    - keywordLocation: /oneOf/2/properties/component-definition/$ref/properties/components/items/$ref/properties/control-implementations/items/$ref/properties/source/$ref/type
      absoluteKeywordLocation: http://csrc.nist.gov/ns/oscal/1.0/1.1.1/oscal-complete-schema.json#/definitions/URIReferenceDatatype/type
      instanceLocation: /component-definition/components/0/control-implementations/0/source
      error: expected string, but got null
    - keywordLocation: /oneOf/2/properties/component-definition/$ref/properties/metadata/$ref/properties/parties/items/properties/links/items/$ref/properties/href/$ref/format
      absoluteKeywordLocation: http://csrc.nist.gov/ns/oscal/1.0/1.1.1/oscal-complete-schema.json#/definitions/URIReferenceDatatype/format
      instanceLocation: /component-definition/metadata/parties/0/links/0/href
      error: '''<https://github.com/defenseunicorns/lula>'' is not valid ''uri-reference'''
      failedValue: <https://github.com/defenseunicorns/lula>
metadata:
    documentPath: src/test/e2e/scenarios/pod-label/oscal-component.yaml
    documentType: component-definition
    documentVersion: 1.1.1
    schemaVersion: 1.1.1
```

mike-winberry: Fixed the component-definition and updated the Assess by upgrading to a temp file and then using that temp file.

brandtkeller (approved):
Functionality is working as I would expect. Appreciate updates / modifications and testing in a manner that won't modify existing artifacts.

mike-winberry:
Also I found the issue that @meganwolf0 was encountering. It has to do with trailing whitespace. It is a current issue with the go yaml implementation https://github.com/go-yaml/yaml/issues/880. But if all trailing spaces are removed it no longer serializes the block/folded strings into double quotes. And is something that could be brought into go-oscal where we trim all empty trailing spaces from lines. So it is a yaml formatting problem. I think that with the right linter it would be caught but an empty space after something like `Group: ` instead of `Group:` breaks the scalar and tells the parser to wrap in double quotes.

brandtkeller:
@mike-winberry I need to think more about edge cases but otherwise that sounds like a great issue for go-OSCAL

mike-winberry:
> @mike-winberry I need to think more about edge cases but otherwise that sounds like a great issue for go-OSCAL

https://github.com/defenseunicorns/go-oscal/issues/167

Have this open right now. With the resulting after removing whitespace.

PR #297, opened 2024-03-07
initial code for scraping resource data fields

### LINKED ISSUE #289: Use data from resource fields for validation
Certain resources contain data in undefined formats (e.g., ConfigMaps, Secrets, perhaps even custom resources), additionally annotations and labels may do so as well. This is difficult for rego to parse as it gets read in as a string conversion of those undefined types. This data would be more usable by the rego validation script if the underlying data type was specified, converted, and given to OPA provider in an expected JSON format. It may also be beneficial if the data can be base64 decoded in cases where applicable.

This issue should look at how to extend the resource type to account for fields, possibly decoding them, in various formats, and providing that data to the OPA provider so that the rego script can access the fields and perform validation on the data.

meganwolf0:
Initial stab at grabbing resource data -> tests are currently a placeholder/in progress; also need some docs.

Closes #289


meganwolf0:
This needs some docs still, but was waiting until the recent changes to docs get merged + all the other code changes that may require some merge deconfliction. But would still love some feedback on the implementation so can re-vector if necessary!

brandtkeller:
@meganwolf0 I don't see any issues with implementation. structure leaves it open for augmenting with some additional ways to get data that is nice. Docs & some explicit criteria in tests would be great.

meganwolf0:
@brandtkeller - Updated the docs! I think based on the chat in sync the tests are sufficient? But let me know if you want more detail there

brandtkeller (approved):
Good addition for the next release.

PR #307, opened 2024-03-14
Lula Version Property

### LINKED ISSUE #294: Lula minimum versioning
Due to the pace of development, new features are being added that might impact how external data (e.g., validation OSCAL) is handled. For instance, two recent issues change the ability to [use named resources](https://github.com/defenseunicorns/lula/issues/249) and [specify outputs](https://github.com/defenseunicorns/lula/issues/265) from the rego provider. For these, the `payload` structure has changed in the back-matter descriptions and those should at least be pinned to minimum Lula versions to ensure that they work as expected.

One possibility could be to augment `target` with a lulaVersion field, such as:
```yaml
target:
   lulaVersion: >= 0.2
```

meganwolf0:
Initial cut at specifying a Lula Version in the "Lula Validation" oscal fields. 
Changes
- New type `Description` for the description to be robust about marshaling that field
- New `IsVersionValid` function
- Updated component.go to do version checking
- Dependency on Hashicorp's go-version library to do the version constraint checking
  -> Since we are using their library, we are beholden to their rules - one of which is that version checks are just automatically denied when you compare a pre-release version to a non-prerelease constraint. I think the logic is that pre-release versions are possibly unstable and shouldn't be used. This creates some issues if building from a branch - which maybe should be discouraged but probably needs to be allowed for development(?). Anyway, needed to create a clean NON_PRERELEASE_VERSION in the Makefile to account for this... If there are other preferred solutions let me know

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,13 +15,14 @@ GIT_COMMIT = $(shell git rev-parse HEAD)
 GIT_SHA    = $(shell git rev-parse --short HEAD)
 GIT_TAG    = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
+NON_PRERELEASE_VERSION ?= $(if $(shell git describe --tags --abbrev=0),$(shell git describe --tags --abbrev=0),"UnknownVersion")
 
 # Go CLI options
 PKG         := ./...
 TAGS        :=
 TESTS       := .
 TESTFLAGS   := -race -v
-LDFLAGS     := -w -s -X 'github.com/defenseunicorns/lula/src/config.CLIVersion=$(GIT_SHA)'
+LDFLAGS     := -w -s -X 'github.com/defenseunicorns/lula/src/config.CLIVersion=$(NON_PRERELEASE_VERSION)'
 GOFLAGS     :=
 CGO_ENABLED ?= 0
 
--- a/Makefile
+++ b/Makefile
@@ -15,14 +15,13 @@ GIT_COMMIT = $(shell git rev-parse HEAD)
 GIT_SHA    = $(shell git rev-parse --short HEAD)
 GIT_TAG    = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
-NON_PRERELEASE_VERSION ?= $(if $(shell git describe --tags --abbrev=0),$(shell git describe --tags --abbrev=0),"UnknownVersion")
 
 # Go CLI options
 PKG         := ./...
 TAGS        :=
 TESTS       := .
 TESTFLAGS   := -race -v
-LDFLAGS     := -w -s -X 'github.com/defenseunicorns/lula/src/config.CLIVersion=$(NON_PRERELEASE_VERSION)'
+LDFLAGS     := -w -s -X 'github.com/defenseunicorns/lula/src/config.CLIVersion=$(CLI_VERSION)'
 GOFLAGS     :=
 CGO_ENABLED ?= 0
 
--- a/docs/version-specification.md
+++ b/docs/version-specification.md
@@ -2,31 +2,32 @@
 In cases where a specific version of Lula is desired, either for typing constraints or desired functionality, a `lulaVersion` property is recognized in the `description` (component-definition.back-matter.resources[_]):
 ```yaml
 - uuid: 88AB3470-B96B-4D7C-BC36-02BF9563C46C
-    title: Lula Validation
-    remarks: >-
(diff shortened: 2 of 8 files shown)


docs/version-specification.md
```diff
@@ -0,0 +1,34 @@
+# Version Specification
+In cases where a specific version of Lula is desired, either for typing constraints or desired functionality, a `lulaVersion` property is recognized in the `description` (component-definition.back-matter.resources[_]):
+```yaml
+- uuid: 88AB3470-B96B-4D7C-BC36-02BF9563C46C
+    title: Lula Validation
+    remarks: >-
+    No outputs in payload
+    description: |
```
brandtkeller: formatting looks off in this example

src/pkg/common/oscal/component.go
```diff
@@ -28,19 +30,41 @@ func BackMatterToMap(backMatter oscalTypes_1_1_2.BackMatter) map[string]types.Va
...
+			err := yaml.Unmarshal([]byte(resource.Description), &description)
 			if err != nil {
 				fmt.Printf("Error marshalling yaml: %s\n", err.Error())
 				return nil
 			}
 
+			// Do version checking here to establish if the version is correct/acceptable
+			var result types.Result
+			var evaluated bool
+			currentVersion := "v0.0.2" //config.CLIVersion
```
brandtkeller: Should this be modified to be that `config.CLIVersion` mention?

src/types/types.go
```diff
@@ -2,6 +2,11 @@ package types
 
 import "errors"
 
+type Description struct {
+	LulaVersion string                 `json:"lulaVersion" yaml:"lulaVersion"`
```
brandtkeller: [no change required] thinking about a TBD standard on what convention we use for naming. should the convention align to OSCAL and `lula-version` or?

Makefile
```diff
@@ -15,13 +15,14 @@ GIT_COMMIT = $(shell git rev-parse HEAD)
 GIT_SHA    = $(shell git rev-parse --short HEAD)
 GIT_TAG    = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
 CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
+NON_PRERELEASE_VERSION ?= $(if $(shell git describe --tags --abbrev=0),$(shell git describe --tags --abbrev=0),"UnknownVersion")
```
brandtkeller: Want to consider implications here that `make build` will produce a binary with the same `lula version` output as that of the binaries we build as part of an official release. It can always be spoofed - but I think this helps quickly identifying a potentially modified binary from one that is straight from a release (If debugging remotely).

`git describe --tags` suffixes the tag name with the number of additional commits on top of the tagged object and the abbreviated object name of the most recent commit. EX: `v0.0.2-57-g33ed319`

Is there a way to set this as the default in the make args and then trim all items right of the semver?

brandtkeller (changes requested):

src/types/types.go
meganwolf0: I was wondering that too - I went with camel case since that's how we did the other fields in the description but if this is going to live in the oscal it does look a little funny to switch conventions

brandtkeller: It also isn't widely enforced on the unmarshall process - so while it appears either works - we should look to migrating `lula-version` and `resource-rule` to the dash-case in implementations (which apprarently doesn't break current camel-case...

meganwolf0: We also use uppercase resourceRule fields too...

docs/version-specification.md
meganwolf0: yeah it is, will fix and choose a better example, I just copy-pasted from my test file and forgot to clean it up

Makefile
meganwolf0: yeah that's a much better solution, I'll modify to do that instead

src/pkg/common/oscal/component.go
meganwolf0: should be fixed

Makefile
meganwolf0: updated, I'll let you take a look again and check if it looks ok to you!

src/types/types.go
brandtkeller: 💯 - it appears that the unmarshall doesn't care (would like to test this). I believe we should take a stance on using `dash case` given we're working within OSCAL - and on first glance doesn't seem to be a breaking change - more of an opinionated implementation.

brandtkeller: Thoughts on taking a stance on dash case and then cleanup/test the other fields in another issue?

brandtkeller: I lied... the mapstructure library is performing the decoding here. So the field name is honored instead of the intuited `yaml:` tag.

`resourceRule` being an outlier.

brandtkeller (approved):
Looks great and sets a good foundation early.

mike-winberry (approved):
Looks good!

PR #309, opened 2024-03-15
Fix: Convert mapstructure decode to yaml unmarshall to type

### LINKED ISSUE #308: Naming conventions for Lula Validation types
Currently there are a variety of naming conventions used in the validations. camel-case, dash-case, etc

Given that validations are intended to be a subset of OSCAL - we should standardize on opinionation of the implmentation. 

## Objectives
- Test whether current unmarshalling from OSCAL validations to data types even cares (is the data still present)
- Migrate to dash case

brandtkeller:
Quick draft for removing the `mapstructure` package and passing `map[string]interface{}` objects around in favor of unmarshalling from yaml directly to types. 

Pros:
- Consistent naming conventions (dash case)
- Leverages a struct type for transient payload data
- Would argue this is a fix for consistent process around yaml - if we ever wanted to support json then the structure is more solidified. 

Cons:
- Does "break" compatibility with current validations

Wanted to raise discussion around the problem space.


meganwolf0 (dismissed):
all the changes look good! the `version-specification.md` doc needs updating with the dash-case for resources, but since that's so small if we're trying to get this over the finish line to release it doesn't seem to me like it should be a blocker

brandtkeller:
> all the changes look good! the `version-specification.md` doc needs updating with the dash-case for resources, but since that's so small if we're trying to get this over the finish line to release it doesn't seem to me like it should be a blocker

Ah good call! I performed the find and replace before the merge from main. Cleaned that up.

meganwolf0 (approved):
lgtm

mike-winberry (approved):

PR #325, opened 2024-03-21
Added command to get resources from validation manifest

### LINKED ISSUE #324: Add command for getting resources directly from a validation statement
Add a command to extract resources from a validation manifest directly.

Input would be a yaml manifest for validation, such as:
```yaml
target:
  provider: opa
  domain: kubernetes
  payload:
    resources:
      - name: deployment
        resource-rule:
          name: monitoring-grafana
          group: apps
          version: v1
          resource: deployments
          namespaces: [monitoring]
    rego: |
      package validate
      
      default validate = false
```

Output would be a json file or json to stdout that represents the resources, which would be inputs to the rego

meganwolf0:
Added `dev get-resources` command to Lula. Need to specify and input file (-f) for the validation manifest and optionally an output file (-o) for the resources json data.

Tested with a local k3d with podinfo deployed and the following validation.yaml input:
```yaml
target:
  provider: opa
  domain: kubernetes
  payload:
    resources:
      - name: deployment
        resource-rule:
          name: podinfo
          group: apps
          version: v1
          resource: deployments
          namespaces: [podinfo]
    rego: |
      package validate
      
      default validate = false
```


meganwolf0:
@brandtkeller - threw this together real quick after our discussion, probably needs more testing and whatnot but would love any feedback on if this is hitting what you were thinking

brandtkeller:
> @brandtkeller - threw this together real quick after our discussion, probably needs more testing and whatnot but would love any feedback on if this is hitting what you were thinking

@meganwolf0 Looks great! E2E test here is likely most applicable. I'd checkout whether `message.JSONValue()` provides any value otherwise this meets the intent.

meganwolf0:
Added some e2e testing and refactored the get-resources command a bit. I also made a change to the resources that get returned and dropped the `metadata.managedFields`. This field seems kind of useless and introduces a lot of lines and potentially noise so I figured it'd be ok to drop. If there's an issue or maybe more discussion is needed I could always revert this and throw it into a separate PR.

src/cmd/dev/get-resources.go
```diff
@@ -0,0 +1,117 @@
...
+	kube "github.com/defenseunicorns/lula/src/pkg/common/kubernetes"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+	"github.com/defenseunicorns/lula/src/types"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+type flags struct {
+	InputFile  string // -f --input-file
+	OutputFile string // -r --result-file
```
brandtkeller: minor copy/paste in the field comments (`-r` vs intended `-o`)

meganwolf0: oops should be fixed!

brandtkeller (approved):

mike-winberry (approved):

PR #339, opened 2024-03-29
feat(dev): #70 add dev validate cmd

### LINKED ISSUE #70: Add CLI option for consuming a validation statement directly
Developer workflow for creating validation payloads for inclusion in OSCAL documentation will take iteration. We can streamline the development process for these validating statements by making the testing of them consume the Lula payload directly instead of via the OSCAL.

mike-winberry:
Closes: #70 
Depends on: #325 

## Whats Changed
- create cmd/dev/validate
- create test/e2e/dev_validation_test

- [x] opa/rego test
- [x] kyverno test
- [x] lint validation or create story
- [x] update docs


meganwolf0:
For me, the nice-to-haves:
- If I could pass an "expected result" such as --expected-result=False and have a return that tells me if the validation output is the same
- If it could return a structured error on the validation.yaml being the wrong format. Right now I think the unmarshaling just fails, it would be nice if that was caught and returned as a structured output
- I'm not sure if it fails if required fields are missing from the yaml? but that would be good as well

mike-winberry:
Linting Resource-rule:
- [x] version and resource are required in resource-rule
- [x] namespaces should be included if name is
- [x] if field is specified, there should also be a name

src/cmd/dev/validate.go
```diff
@@ -0,0 +1,145 @@
...
+	"strings"
+
+	"github.com/defenseunicorns/go-oscal/src/pkg/utils"
+	"github.com/defenseunicorns/lula/src/cmd/validate"
+	"github.com/defenseunicorns/lula/src/config"
+	"github.com/defenseunicorns/lula/src/pkg/common"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+	"github.com/defenseunicorns/lula/src/types"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
```
meganwolf0, line 17: Seems like we using a mix of this module and gopkg.in/yaml.v3 - should we consolidate?

src/cmd/dev/validate.go
```diff
@@ -0,0 +1,157 @@
...
+	validateCmd := &cobra.Command{
+		Use:   "validate",
+		Short: "Run an individual Lula validation.",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			config.SkipLogFile = true
+		},
+		Long:    "Run an individual Lula validation for quick testing and debugging of a Lula Validation. This command is intended for development purposes only.",
+		Example: validateHelp,
+		Run: func(cmd *cobra.Command, args []string) {
+			spinner := message.NewProgressSpinner("Validating %s", validateOpts.InputFile)
```
brandtkeller: Between here and below - if you execute `lula dev validate` and do not pass any flags or stdin input - Lula will just sit with a spinner that says `Validating 0`. Edge-case but is there anything we want to do here? Both from the naming perspective and from the error handling?

mike-winberry: Yeah Ill take a look at that and ensure it exits properly!

mike-winberry: I added in a timeout and updated the help command with instructions as reading from stdin's default behavior is to hang until EOF. Also updated the message when it hangs to say what it is doing.

meganwolf0 (dismissed):

src/cmd/dev/validate.go
mike-winberry: I think we will need a new issue for this, k8s is required for kyverno stuff, but doesn't handle most json and other cases. So right now neither will properly handle unmarshalling and marshalling for all cases.

mike-winberry: created: https://github.com/defenseunicorns/lula/issues/362

CloudBeard (dismissed):
command and flags seem to work for me, didnt see any issues.

src/cmd/dev/validate.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/defenseunicorns/lula/src/types"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+)
+
+const STDIN = "0"
+const NO_TIMEOUT = -1
+const DEFAULT_TIMEOUT = 1
+
+var validateHelp = `
```
brandtkeller, line 24: Nitpick - I don't see an alias for `lula dev validation` as the help block indicates.

mike-winberry: I think that is due to it needing to process stdin. I could change to require something like `lula dev validate -f stdin` but that breaks the "unix" convention, don't know if that matters to us.

mike-winberry:
```
                                                                                                                                                              
☁  lula [298-validations-in-a-separate-manifest] ./bin/lula dev validate --help
Run an individual Lula validation for quick testing and debugging of a Lula Validation. This command is intended for development purposes only.

Usage:
  lula dev validate [flags]

Examples:

To run validations using a lula validation manifest:
        lula dev validate -f <path to manifest>
To run validations using stdin:
        cat <path to manifest> | lula dev validation
To hang indefinitely for stdin:
        lula dev validation -t -1
To hang for timeout of 5 seconds:
        lula dev validation -t 5


Flags:
  -e, --expected-result      the expected result of the validation (-e=false for failing result) (default true)
  -h, --help                 help for validate
  -f, --input-file string    the path to a validation manifest file (default "0")
  -o, --output-file string   the path to write the validation with results
  -t, --timeout int          the timeout for stdin (in seconds, -1 for no timeout) (default 1)

Global Flags:
  -l, --log-level string   Log level when running Lula. Valid options are: warn, info, debug, trace (default "info")
  ```

mike-winberry: The help command still works as intended.

brandtkeller: That makes sense - more highlighting this `validateHelp` string contains `lula dev validation` which isn't a real command (doesn't work) because the command is `lula dev validate`.

mike-winberry: nvm I actually fixed it, oh lol Ill fix that too!

brandtkeller (approved):

CloudBeard (approved):

PR #354, opened 2024-04-02
ADR for validation yaml

### LINKED ISSUE #332: Refactor domain and resources
The current validation structure has a somewhat confusing use of `domain` and `resources`, where they appear separate in the definition, but are coupled in practice - e.g.,
```yaml
target:
  provider: opa
  domain: kubernetes
  payload:
    resources:
      - {list of resources}
```

This flows into the code organization - e.g., getting the resources is implemented redundantly inside the provider modules.

The structure of both the validation yaml and implementation in code should probably be refactored to reduce redundant usage and enhance clarity.

meganwolf0:
This is a **really** rough first outline at what some possible options could look like for the validation yaml when we consider:
- Domain / Provider decoupling
- Remote validations
- Logical validations
- Placeholder validations

I think it might make sense to have multiple specifications for the yaml, I'm not sure if it's an anti-pattern to do a bunch of case sensitive switching on field structure, but it feels a little icky.


adr/0007-validation-yaml-format.md
```diff
@@ -0,0 +1,169 @@
...
+
+```yaml
+lula-version: "1.0"                           # Optional (maintains backward compatilibity)
+metadata:                                     # Required
+  title: "title here"                         # Optional (short description to use in output of validations could be useful)
+  type: local                                 # Required (enum:[local, remote, admin]??)
+  workflow: healthcheck                       # Required (enum: [satisfaction, healthcheck]) - basically this indicates some way that this validation is mapped to results
+spec:
+  domain: kubernetes                          # Required (enum:[kubernetes])
+    domain-spec:                              # Required (the spec itself can be dependent on the domain chosen, so can be flexible to new domains like infra)
```
brandtkeller: What do `domain-spec` and `provider-spec` offer us logically? Looking to understand more of the intent here.

meganwolf0: Did I actually screw up that part? I think for it to be valid yaml the domain and provider would be like
```yaml
domain:
  type: kubernetes
  domain-spec: ... 
```

I was just wanting some way to indicate those are the swappable pieces depending on what the domain or provider type is, for readability and for the code. I figured reusing "spec" would also be overloaded, but maybe that's fine for this. From a linting perspective I thought it would be easier to just have fully contained structures there that could be validated based on the domain/provider types.

adr/0007-validation-yaml-format.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Example for passthrough or placeholder validation: 
+
+```yaml
+target:
+  domain: 
+    type: passthrough                         # Required (enum:[kubernetes, passthrough])
+    passthrough-spec:
+      evaluation: INTERVIEW                   # Required: INTERVIEW | EXAMINE
```
CloudBeard: In the comment would it be good to call out TEST as well? That would include all 3 options. There may be cases where it could be automated once someone does a thing manually to complete a setup.

meganwolf0:
I'm wondering if the domain.type and provider.type is redundant... that information is effectively already given in the x_spec field... so really it's more there from like a decorator perspective but in the interest of trimming fat perhaps we could do-away with it? I guess the one benefit of having it is if, for example, someone had specs for multiple providers, the "type" would point to which should actually be used. Again, perhaps not necessary though since you could probably determine logic for handling that scenario in the tool.

adr/0007-validation-yaml-format.md
meganwolf0: Yeah that's a good point, I can update that one

CloudBeard:
> I'm wondering if the domain.type and provider.type is redundant... that information is effectively already given in the x_spec field... so really it's more there from like a decorator perspective but in the interest of trimming fat perhaps we could do-away with it? I guess the one benefit of having it is if, for example, someone had specs for multiple providers, the "type" would point to which should actually be used. Again, perhaps not necessary though since you could probably determine logic for handling that scenario in the tool.

@meganwolf0 have you thought about when we support AWS/CSPs and Terraform etc? I think the domain/type field has more merit there. 

Also wanted to call out that I like what you've got marked as required and optional really do line up to support other domains. Since  AWS/etc may not need that but it may need/could use others.

adr/0007-validation-yaml-format.md
```diff
@@ -0,0 +1,207 @@
...
+                      # Note the `~.` modifier, it means we want to iterate over array elements in descendants
+                      ~.containers:
+                        image:
+                          # Check that an image tag is present
+                          (contains(@, ':')): true
+                          # Check that the image tag is not `:latest`
+                          (ends_with(@, ':latest')): false
+```
+
+### Proposal
```
brandtkeller: Do we want to include here an additional parent element that captures defining a list of validations?

adr/0007-validation-yaml-format.md
```diff
@@ -0,0 +1,207 @@
...
+                      # Note the `~.` modifier, it means we want to iterate over array elements in descendants
+                      ~.containers:
+                        image:
+                          # Check that an image tag is present
+                          (contains(@, ':')): true
+                          # Check that the image tag is not `:latest`
+                          (ends_with(@, ':latest')): false                       
+```
+
+Example for passthrough or placeholder validation:
```
brandtkeller: This helps add some context for the `x_spec` field but may contribute to some historical confusion over time. IE if we implement passthrough/placeholder (which feels like a separate issue) but use different language or fields - search might confuse people. 

Rather i'd be more interested about potential type structures and what extending this looks like.

adr/0007-validation-yaml-format.md
meganwolf0: I think having the yaml documents in a single file but separated by `---` would allow them to be ingested by the yaml libs as an array of documents, so I'm not sure we need an overarching structure to contain the list of things - unless I'm missing something else about that workflow

brandtkeller: Just thinking about the processing piece for inclusion in a map.

In order to unmarshall a single document that contains many validations separated by `---` we'll have to parse the separate ourselves (unless that is a builtin somewhere?). 

versus having a parent element that could be unmarshalled wholesale. 


Not having a parent element would mean more alignment between `lula dev` commands and a single validation.

adr/0007-validation-artifact-format.md
```diff
@@ -0,0 +1,165 @@
...
+  name: "title here"                          # Optional (short description to use in output of validations could be useful)
+target:
+  domain: 
+    type: kubernetes                          # Required (enum:[kubernetes, api])
+    kubernetes-spec:                          # Optional
+      resources:                                  
+      - name: podsvt                          # Required 
+        resource-rule:                        # Required
+          version: v1                         # Required
+          kind: pods                          # Required (formerly "resource" but "kind" seems to make more sense in a k8s context)
```
brandtkeller: Want to open discussion here about this change `resource` -> `kind`.

Given my understanding of [resource vs kind](https://github.com/kubernetes/community/blob/7f3f3205448a8acfdff4f1ddad81364709ae9b71/contributors/devel/sig-architecture/api-conventions.md#resources) and the explicit definition of the `schema.GroupVersionResource` object that we are creating for retrieving kubernetes data dynamically - I am unsure how large of a change this would be to swap. 

Would need to dive further on the delta between `GVR` and `GVK` here - but there potentially could be support for kind without introducing any breaking change?

adr/0007-validation-artifact-format.md
```diff
@@ -0,0 +1,165 @@
...
+            jsonpath:                         # Required
+            type:                             # Optional 
+            base64:                           # Optional 
+      wait:                                   # Optional 
+        condition: Ready                      # Optional 
+        kind: pod/test-pod-wait               # Optional 
+        namespace: validation-test            # Optional 
+        timeout: 30s                          # Optional 
+
+  provider:
```
mike-winberry: Looks good! My only thoughts are that there is some redundancy in the field names ie:
```yaml
type: opa
// We already have the type so could this just be "spec"?
opa-spec:
```
Same in the kyverno
```yaml
type: kyverno
kyverno-spec:
  kyverno:
```

Is there a way we can reduce the redundancy and more align the types since we already declare the `type: opa/kyverno`

mike-winberry (changes requested):

adr/0007-validation-artifact-format.md
mike-winberry: Not really requesting changes as it isn't a major issue just wanted to highlight for discussion before we merge this and #332

mike-winberry (dismissed):

adr/0007-validation-artifact-format.md
brandtkeller: Discussed as a potential future investigation and feature

adr/0007-validation-artifact-format.md
brandtkeller: Discussed the complexity involved with making this changed and agreed that the current `x-spec` option provided more readability around struct use.

CloudBeard (approved):
Late to the show. Everything looks good/makes sense to me!

brandtkeller (approved):

PR #357, opened 2024-04-04
feat:  #298 Add the ability to pull validations from external sources (currently only backmatter)

### LINKED ISSUE #298: Validations in a separate manifest
As both an enabler of re-use and improving developer experience for the artifacts that Lula needs to assess and map against OSCAL artifacts - having the ability to support remote references could serve this need.

Lula validations with an understood remote reference could improve the developer experience of the validations, as well as clean up the use of multiline markup fields.

mike-winberry:
## Description
In order to make lula validations more accessible and reusable, the ability to add validations from external sources should be implemented. 

- feat: create type `types.LulaValidationMap`
- feat: add ability to run multiple validations from a single yaml delimited with `---`
- feat: add ability to target single validation in a yaml with multiple validations
- feat: add ability to use remote vaidations
- docs: Update oscal-validation-links with info and examples on external validations
- [x] feat: create resource-fragment wildcard "token" for running all validations in a multi validation file

## Related issue
#198 

## Type of change
- [x] feature

## additional tasks
- [x] Update docs
- [x] testing


docs/oscal-validation-links.md
```diff
@@ -66,5 +75,52 @@ links:
...
+### Checksums
+- A checksum may be provided in the href using the suffix `@<checksum>` 
+- Supports `sha1`, `sha256`, `sha512`, `md5`
+```yaml
+links:
+  - href: https://example.com/validation.yaml@0123456789abcdef
+    rel: lula
+```
+
+### Validation Maps
```
brandtkeller: The structure of the validation map is the only item I see that we should come to a consensus on.

Would like weigh this implementation versus a yaml file consisting of many validations that is delimited by the standard `---` and then possibly placing `uuid` in the metadata field. 

Pros and Cons to each - but something that we might want to think about.

meganwolf0: Personally I like the look of multiple files delimited by `---` better, plus they're slightly easier to compose if you are just interested in evaluating all of them because then you wouldn't actually need to generate a uuid - which is nice because there's already a lot of those

mike-winberry: I can see pros of both, I also see some major limitations in scalability long term as delimited files are essentially just multiple documents in one, making them non-indexable, and not really that great for an "ecosystem" of validations, but if the scope is to have it initially just apply all validations to a single control id, then I guess it makes sense, though either way, I think we may need to push this off of the milestone, as it is pretty outside initial scope. 

I am going to branch this PR and experiment. And try and discuss at sync/office hours, wether it makes sense to remove the multiple file validation for now and bump a new story, or just push this story to the next milestone.

mike-winberry: Have a PR open to merge into this branch if desired with the changes to handle running multiple validations from a file with `---` delimited validations and targeting a single validation within a `---` delimited validation file. 

#366

meganwolf0 (dismissed):

meganwolf0 (dismissed):

src/test/e2e/scenarios/remote-validations/validation.kyverno.yaml
```diff
@@ -0,0 +1,31 @@
+lula-version: ">= v0.1.0"
```
brandtkeller: nitpick - Unsure on the accuracy of the version constraint here - as this won't actually work with the `v0.1.0` release.

docs/oscal-validation-links.md
```diff
@@ -66,5 +75,58 @@ links:
...
+
+## Importing Validations
+In addition to storing validaitons in the `BackMatter`, `links` may be used to fetch resources external to the `component-definition`.
+
+### Local Validations
+- must be prefixed with `file:`
+- `file:` must be a relative path to the `component-definition` or an absolute path
+```yaml
+links:
+  - href: file:./validation.yaml
```
brandtkeller, line 90: Still reading [rfc-3986](https://www.rfc-editor.org/rfc/rfc3986#section-1.1) but when is `file://` required vs `file:`?

mike-winberry: If there is authority, ie host name like local host. From my understanding of
https://www.rfc-editor.org/rfc/rfc3986#section-3.3

mike-winberry: could be `file:///`, `file://localhost`, or `file://./`

brandtkeller (approved):

meganwolf0 (approved):

PR #358, opened 2024-04-04
Initial refactor for domains and providers

### LINKED ISSUE #332: Refactor domain and resources

meganwolf0:
Some initial stabs at refactoring the code to support domains and providers in a (hopefully) more extensible and modular way


src/pkg/common/types.go
```diff
@@ -0,0 +1,37 @@
+package common
+
+import (
+	"github.com/defenseunicorns/lula/src/pkg/domains/api"
+	kube "github.com/defenseunicorns/lula/src/pkg/domains/kubernetes"
+	"github.com/defenseunicorns/lula/src/pkg/providers/kyverno"
+	"github.com/defenseunicorns/lula/src/pkg/providers/opa"
+)
+
+// YAML data structures for ingesting validation data
```
brandtkeller: Does any of this really lock us in to yaml? wondering what would need to be modified in order to support JSON and if that would change the naming here?

src/types/types.go
```diff
@@ -1,100 +1,46 @@
 package types
 
-import (
-	"errors"
-	kjson "github.com/kyverno/kyverno-json/pkg/apis/policy/v1alpha1"
+type LulaValidationType string
+
+const (
+	LulaValidationTypeNormal  LulaValidationType = "Lula Validation"
```
brandtkeller, line 6: let's discuss validation "types" more. Curious if that is still within the boundary of validations or not.

src/pkg/common/types.go
meganwolf0: No I think you're right, this probably shouldn't be called "yaml" - I was trying to figure out how to delineate between the pieces that would be like I/O vs the lula-internal structures (which I tried to bucket into types/types.go) but happy to rename to what makes sense there

src/pkg/providers/kyverno/types.go
```diff
@@ -0,0 +1,34 @@
...
+func (k KyvernoProvider) Evaluate(resources types.DomainResources) (types.Result, error) {
+	results, err := GetValidatedAssets(k.Context, k.Spec.Kyverno, resources, k.Spec.Output)
+	if err != nil {
+		return types.Result{}, err
+	}
+	return results, nil
+}
+
+type KyvernoSpec struct {
+	Kyverno *kjson.ValidatingPolicy `json:"kyverno" yaml:"kyverno"`
```
brandtkeller: No change required - thinking about if `kyverno` is the most intuitive option here. 

`opa-spec` has `rego` 

Should `kyverno-spec` have `policy` ?

brandtkeller (changes requested):
Needs `docs/` files updated.

mike-winberry (commented):
Looks good, just have one item for discussion on #354

mike-winberry (approved):

brandtkeller (approved):

PR #361, opened 2024-04-10
feat: Establish workflow for lula validation identifiers

### LINKED ISSUE #226: Improve Lula validation identifier
Currently links and resources require very specific naming `Lula Validation` in order for lula to know and execute validations in the OSCAL. 

Evaluate existing fields for `implemented-requirements` and `backmatter.resources` for a better identifier that is less likely to incur errors due to typos.

brandtkeller:
Closes #226 

Improving the validation identifiers to improve the developer experience and reduce potential errors due to miss identifying a link. 

## Changes 
- Provide better separation between the OSCAL and Lula validations
  - `BackMatterToMap()` now strictly provides a map[string]string object for which identified validations will then be unmarshalled and placed in the `validationMap`. 
    - This removes the need to place any identifier in the `back-matter` validation
    - This also provides a greater separation from OSCAL functionality and Lula Validation functionality
- Currently backwards compatible - this can be discussed and removed
- Establishes the use of `rel` for which we _could_ expand to include validation "types"
- `validationMap` serves future external processing - such that when an external document is unmarshalled, we can place 1 -> N validations into the map rather than having to open the same file multiple times for different validations (reuse).
- Move the unmarshall and validation of a Validation from a String (back-matter.resource[_].description) to the common package to delineate that functionality from OSCAL functionality.

### DIFF
--- a/src/cmd/validate/validate.go
+++ b/src/cmd/validate/validate.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/defenseunicorns/go-oscal/src/pkg/uuid"
 	oscalTypes_1_1_2 "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-2"
+	"github.com/defenseunicorns/lula/src/pkg/common"
 	"github.com/defenseunicorns/lula/src/pkg/common/oscal"
 	"github.com/defenseunicorns/lula/src/pkg/message"
 	"github.com/defenseunicorns/lula/src/types"
@@ -98,7 +99,7 @@ func ValidateOnPath(path string) (findingMap map[string]oscalTypes_1_1_2.Finding
 
 	_, err = os.Stat(path)
 	if os.IsNotExist(err) {
-		return findingMap, observations, fmt.Errorf("Path: %v does not exist - unable to digest document\n", path)
+		return findingMap, observations, fmt.Errorf("path: %v does not exist - unable to digest document", path)
 	}
 	data, err := os.ReadFile(path)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


meganwolf0 (dismissed):

src/cmd/validate/validate.go
```diff
@@ -122,8 +124,12 @@ func ValidateOnPath(path string) (findingMap map[string]oscalTypes_1_1_2.Finding
 // It will perform a validation and add data to a referenced report object
 func ValidateOnCompDef(compDef oscalTypes_1_1_2.ComponentDefinition) (map[string]oscalTypes_1_1_2.Finding, []oscalTypes_1_1_2.Observation, error) {
 
-	// Populate a map[uuid]Validation into the validations
-	validations := oscal.BackMatterToMap(*compDef.BackMatter)
+	// Populate a map[uuid]string for back-matter.resources
+	// This is pre-poulated for on-demand use when referencing the back-matter from a link
+	backMatterMap := oscal.BackMatterToMap(*compDef.BackMatter)
```
mike-winberry, line 129: Just curious why the seperation here? is there something I am missing as to why we want to maintain two maps, rather that just eating the cost of processing the backmatter resources and getting the LulaValidations right off the bat?

mike-winberry (changes requested):

src/cmd/validate/validate.go
mike-winberry: Other than that question all looks good!

brandtkeller: Great question and a design choice here that we can certainly dissect or revert.

Initial goal was to remove the need for any identifier in the back-matter resources. It creates additional requirements to remember and more room for error.

I see two options (maybe there are more?):
1. In the proposed logic - have the `BackMatterToMap` function simply return uuid/description (string) map for ALL back-matter resources. Then only attempt to unmarshall to a Validation object when the link contains `rel: lula`.
2. Have functionality that checks the validity of the object created and attempt to unmarshall all back-matter resources - only keeping the valid ones. 

Open to feedback.

mike-winberry: That makes sense! Im good. I think it can be addressed later if it needs to, I don't really see it being an issue though. Just was missing that connection between removing the id in the backmatter.

mike-winberry (approved):

meganwolf0 (approved):

CloudBeard (approved):

PR #363, opened 2024-04-10
chore: #46 add issue and pr templates

### LINKED ISSUE #46: Add Issue and Pull Request templates
Why do we need this?

To help provide shared terminology and a consistent format for people to submit questions, bug fixes, feature requests, code/documentation contributions, etc.

Suggested solution:

We should try to re-use as much as we can from what Zarf currently uses for [issue templates](https://github.com/defenseunicorns/zarf/tree/main/.github/ISSUE_TEMPLATE) and [pull request templates](https://github.com/defenseunicorns/zarf/blob/main/.github/pull_request_template.md) to help stay consistent across projects.

mike-winberry:
- Moved Issue and pr template from go-oscal
- Update issue and pr templates with lula links instead of go-oscal

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: possible-bug
+assignees: ''
+---
+
+### Environment
+Device and OS:
+App version:
+Kubernetes distro being used:
+Other:
+
+### Steps to reproduce
+1.
+
+### Expected result
+
+### Actual Result
+
+### Visual Proof (screenshots, videos, text, etc)
+
+### Severity/Priority
+
+### Additional Context
+Add any other context or screenshots about the technical debt here.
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+about: Create a report to help us improve
+title: ''
+labels: possible-bug
+assignees: ''
+---
+
+### Environment
+Device and OS:
+App version:
+Kubernetes distro being used:
```
CloudBeard: Is this for the provider being used in validations?

mike-winberry: No but that's a good idea! I'll add that as a field, this is just the basic co. Template

brandtkeller: Version of kubernetes would be really valuable for reproducing errors. We've already seen some functionality that doesn't work on versions of k8s that is now EOL.

CloudBeard (approved):

brandtkeller (approved):

Issue #247, opened 2024-02-06
Discovery for Report/Visual Creation


CloudBeard:
# What
Need for outputting the data obtained from the OSCAL and validations to create high level reports and possible graphical high level representations.

# How
Taking something in theory like the Compliance MD Generator's logic and grabbing those results from OSCAL and Validations using Lula would enable us to create these reports. That logic would also be useful for GUI integration at a later date.

Visualization piece of an SVG bar graph. Possibly a GitHub action or custom piece to take the data from Lula to create this image?

## Note
This is not meant to be a replacement for a GRC tool or proper visualization with a GUI of some sort. This is for a quick and easy way to showcase controls for groups like Delivery or anyone who just needs the quick metrics.

brandtkeller:
Relates to #223 - My opinion is that we should develop some of the underlying foundation for provenance/accuracy of the data we are looking to report before attempting to visualize.

CloudBeard:
100% agree, need the data before reporting on it.

Issue #223, opened 2024-01-19
Reporting of number of controls met by component definitions


brandtkeller:
### Is your feature request related to a problem? Please describe.

Ability to report on the number of controls present in a component definition - filtered by control-implementation source and framework. 

### Describe the solution you'd like

- **Given** a component definition is present
- **When** `lula` report is executed
- **Then** a report logged to stdout to capture number of controls based on the above filter

### Additional context

This would pave the way for future reporting on percentage of controls (when Lula can properly process both catalogs and profiles).

### Expected Deliverable

- [ ] Report Command added
- [ ] Ability to ingest component-definition files (error otherwise)
- [ ] Aggregate number of controls by control-implementation source and framework
- [ ] Log report to stdout
- [ ] Tests
- [ ] Docs

brandtkeller:
```mermaid
flowchart TD
    A[Report] -->|default/specified OSCAL files| B(Build Report)
    B --> C{Check for existence of models & Collect/identify Catalog}
    C -->|Components Exist| D[Component Report]
    C -->|SSP Exists| E[System Report]
    C -->|Assessment Results Exist| F[Assessment Report]
```

CloudBeard:
That makes perfect sense ^

CloudBeard:
Need to swap my tests back around before re-review

PR #372, opened 2024-04-16
feat: initial poc for report

CloudBeard:
## Description

Initial POC for report command. Currently only reports controls from component definition. Outputs json by default but also supports yaml. Support the ability to create output file if needed and stdout. This allows for broader use of the report but also ability to send JSON/YAML into a processor for visualization.

Currently can create a detailed and/or summary report based on needs.

## Related Issue

Relates to #223 and #247 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/report/common.go
+++ b/src/cmd/report/common.go
@@ -1,0 +1,16 @@
+package report
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var reportCmd = &cobra.Command{
+	Use:     "report",
+	Aliases: []string{"r"},
+	Short:   "Collection of report commands to make assessor life easier",
+}
+
+// Include adds the tools command to the root command.
+func Include(rootCmd *cobra.Command) {
+	rootCmd.AddCommand(reportCmd)
+}
--- a/src/cmd/report/component-definition.go
+++ b/src/cmd/report/component-definition.go
@@ -1,17 +1,12 @@
 package report
 
 import (
-	// "bytes"
 	"errors"
 	"fmt"
 	"os"
 	"encoding/json"
-	// "strings"
-	// "time"
 
-	// "github.com/defenseunicorns/go-oscal/src/pkg/uuid"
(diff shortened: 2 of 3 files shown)


CloudBeard:
Still need to add functionality for grabbing all controls from source profile to get a percentage mapped.

All other report types SSP/SAP/SAR would be added once we have those to pull data from.

Still need to add tests.

brandtkeller:
For posterity - we had discussed this being blocked/dependent on the generation work. 

After re-review, I believe that is short-sighted on my part. While not untrue for the code - I think there is more that we could be doing to effectively identify the problem. 

I appreciate what you have done here - I think what we lack currently is what the intent of Lula Reporting really entails. You have created some flexibility wherein we can report on different models or targets - which may very well be a part of what we need.

To make this more actionable - I believe we need a higher-level picture of what is valuable and that we want to accomplish with the `lula report`. Possibly a design session to brainstorm and create agreement on direction? Open to thoughts and feedback.

CloudBeard:
I agree on the design session, I didn't want to get too in-depth with what is being reported and how its gotten. How much is by default, do we go with a detailed/summary, do we go with a way to add more or remove from those.

I can see the need to provide tons of details. How many controls total, how many in X family, tech vs admin, new term portable vs non-portable (if you piece mill it or use another CSP what do you lose/keep), various frameworks/ILs, how many are validated, how many failed, and tons of correlations of each of these.

All of that for every layer component definition, SSP, SAP/SAR, POAM

Figuring out what is the most important to start with and what is needed to do it. I think Daniel will have a lot of awesome insights on what could be initially needed to show.

I only think its "blocked" in the sense we just need a working OSCAL with some example data of each to report against. I believe we could run it along side generate functionality. Possibly use report to test the correctness of generate's output.

brandtkeller:
@CloudBeard Do we still need this Draft given #599 ?

CloudBeard:
> @CloudBeard Do we still need this Draft given #599 ?

Nope ill close this one. I used a mix of your draft and this draft to get started.

PR #384, opened 2024-04-23
feat!: #367 compiling external/remote validations

### LINKED ISSUE #367: Compiling External/Remote Validations
In order to support disconnected environments, the ability to "compile" validations into a component-definition, should be created in order to support the portability of component-definitions that rely on remote/external validations. 

- **Given** a component-definition with links to external validations
- **When** running the command `lula compile` (wip) with a component-definition file
- **Then** the validations that exist outside the backmatter are pulled and added to backmatter.resources and the links are updated to point to the backmatter instead of the external validation. 

### Additional context
- depends on #357

mike-winberry:
## Description
Create compile command so that component-definitions with external lula validations can be portable in disconnected environments. 

## Related Issue
#367 

## Type of change

- [x] New feature

## Checklist before merging


brandtkeller:
What are your thoughts around location of the command? 

CLI root vs as a subcommand elsewhere?

mike-winberry:
> What are your thoughts around location of the command?
> 
> CLI root vs as a subcommand elsewhere?

I was having similar thoughts. I feel like it could fit under a `tools` path or just under `dev` as well. Wasn't sure as it will likely expand to include compiling more than just validations which is why I started at root, but am not sold entirely. What are your thoughts?

brandtkeller:
> > What are your thoughts around location of the command?
> > CLI root vs as a subcommand elsewhere?
> 
> I was having similar thoughts. I feel like it could fit under a `tools` path or just under `dev` as well. Wasn't sure as it will likely expand to include compiling more than just validations which is why I started at root, but am not sold entirely. What are your thoughts?

It is not strictly tied to the "core" capability surrounding what Lula performs - rather "helps" with managing validations. 

Maybe a `tools` candidate?

src/pkg/common/compilation/resource-store.go
```diff
@@ -0,0 +1,143 @@
...
+	validationBytesArr := bytes.Split(validationBytes, []byte(common.YAML_DELIMITER))
+	isSingleValidation := len(validationBytesArr) == 1
+
+	for _, validationBytes := range validationBytesArr {
+		var validation common.Validation
+		if err = validation.UnmarshalYaml(validationBytes); err != nil {
+			return nil, err
+		}
+
+		resource, err := validation.ToResource()
```
brandtkeller: Do we need to evaluate changing optional non-primitives to pointers as we did in `go-oscal`?

mike-winberry: idk that it is necessary right now but I definitely think it is worth consideration. Ill create a discussion issue.

mike-winberry: https://github.com/defenseunicorns/lula/issues/388

brandtkeller:
Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?

src/pkg/common/compilation/compilation.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+	}
+	allFetched := resourceMap.AllFetched()
+	if compDef.BackMatter != nil && compDef.BackMatter.Resources != nil {
+		existingResources := *compDef.BackMatter.Resources
+		existingResources = append(existingResources, allFetched...)
+		compDef.BackMatter.Resources = &existingResources
+	} else {
+		compDef.BackMatter = &oscalTypes_1_1_2.BackMatter{
```
brandtkeller: are there any fields within the metadata that we should be updating? thinking about items like `last-modified`.

mike-winberry:
> Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?

I would may be shift to compose/decompose may be? I think that could be appropriate. I can't think of much else that would make sense right now, ill think on it.

brandtkeller:
> > Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?
> 
> I would may be shift to compose/decompose may be? I think that could be appropriate. I can't think of much else that would make sense right now, ill think on it.

agreed! or thinking about some verb that captures both scenarios (compose may still do this) such as tooling like `base64` where `base64` and `base64 -e` both indicate encryption while `base64 -d` specifies decryption.

mike-winberry:
®

> > > Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?
> > 
> > 
> > I would may be shift to compose/decompose may be? I think that could be appropriate. I can't think of much else that would make sense right now, ill think on it.
> 
> agreed! or thinking about some verb that captures both scenarios (compose may still do this) such as tooling like `base64` where `base64` and `base64 -e` both indicate encryption while `base64 -d` specifies decryption.

I guess I am not really seeing the relation to base64, as base64 is the program name with -d -e being decryption and encryption respectively, which would still translate to two separate commands using cobra such as `pack/unpack` `compose/decompose`, `zip/unzip`, `embed/extract`. If we want to get really specific, since this isn't a `lula` command, should it be nested under oscal `lula tools oscal -c/-d`? I guess I don't really see the benefit of combining the actions, but I may be missing the point, or overthinking it.

brandtkeller:
> ®
> 
> > > > Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?
> > > 
> > > 
> > > I would may be shift to compose/decompose may be? I think that could be appropriate. I can't think of much else that would make sense right now, ill think on it.
> > 
> > 
> > agreed! or thinking about some verb that captures both scenarios (compose may still do this) such as tooling like `base64` where `base64` and `base64 -e` both indicate encryption while `base64 -d` specifies decryption.
> 
> I guess I am not really seeing the relation to base64, as base64 is the program name with -d -e being decryption and encryption respectively, which would still translate to two separate commands using cobra such as `pack/unpack` `compose/decompose`, `zip/unzip`, `embed/extract`. If we want to get really specific, since this isn't a `lula` command, should it be nested under oscal `lula tools oscal -c/-d`? I guess I don't really see the benefit of combining the actions, but I may be missing the point, or overthinking it.

yeah maybe not the greatest example and given the current `lula tools <command>` execution it may not be a huge deal - wouldn't stew on it. If in the future we have more automation specific to OSCAL then I don't hate the idea of nesting `compose` under `lula tools oscal` .

mike-winberry:
> > ®
> > > > > Would you change anything about the verbiage if we wanted this activity to work in both directions? IE compile and decompile?
> > > > 
> > > > 
> > > > I would may be shift to compose/decompose may be? I think that could be appropriate. I can't think of much else that would make sense right now, ill think on it.
> > > 
> > > 
> > > agreed! or thinking about some verb that captures both scenarios (compose may still do this) such as tooling like `base64` where `base64` and `base64 -e` both indicate encryption while `base64 -d` specifies decryption.
> > 
> > 
> > I guess I am not really seeing the relation to base64, as base64 is the program name with -d -e being decryption and encryption respectively, which would still translate to two separate commands using cobra such as `pack/unpack` `compose/decompose`, `zip/unzip`, `embed/extract`. If we want to get really specific, since this isn't a `lula` command, should it be nested under oscal `lula tools oscal -c/-d`? I guess I don't really see the benefit of combining the actions, but I may be missing the point, or overthinking it.
> 
> yeah maybe not the greatest example and given the current `lula tools <command>` execution it may not be a huge deal - wouldn't stew on it. If in the future we have more automation specific to OSCAL then I don't hate the idea of nesting `compose` under `lula tools oscal` .

Makes sense, Ill make the switch to `compose`, since that has the best antonym I have been able to find with `decompose` as the others don't really make sense for what it is or just don't have a good opposite that is intuitive like `resolve` or `aggregate`.

meganwolf0 (approved):
Tested composition pulling from various sources and worked as expected! lgtm

brandtkeller (approved):
Verified the new command location and the changes for updating the `last-modified` field are working as expected.

PR #394, opened 2024-04-26
Dev validate with optional resources file input

### LINKED ISSUE #385: Add input resources.json to dev validate command
This may not bucket in `dev validate` and may possibly be a separate command, but it would be beneficial to have the ability to pass in resource.json to the command to execute expected validations. This is particularly useful in testing of policy, where you'd have a different expected outcome for different resource test cases. It's not realistic to have to modify the cluster to match the scenario, so it would be easier to just change the resource json. This would also support having some documentation of test cases ran against each validation.

- **Given** I have different resource json that map to different validation outcomes (True/False)
- **When** I want to test a Lula validation
- **Then** I'm able to choose those resources + pass an expected validation outcome

None

### Additional context
None

meganwolf0:
## Description

Added optional -r or --resources-file with expected input of *.json file. This will pass the resources to the validate function as an option to use those "static" resources vs the runtime "dynamic" ones pulled by the domain function GetResources.

Figured if other options were desired for the "Validate" function, those could be extended in the `lulaValidationOptions`

## Related Issue

Relates to #385 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/dev/validate.go
+++ b/src/cmd/dev/validate.go
@@ -24,6 +24,8 @@ const DEFAULT_TIMEOUT = 1
 var validateHelp = `
 To run validations using a lula validation manifest:
 	lula dev validate -f <path to manifest>
+To run validations using a custom resources file:
+	lula dev validate -f <path to manifest> -r <path to resources file>
 To run validations using stdin:
 	cat <path to manifest> | lula dev validate
 To hang indefinitely for stdin:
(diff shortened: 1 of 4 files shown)


brandtkeller (approved):
Ran a few additional scenarios through using the resources flag with expected results. Will file an issue to get the `dev` command docs created.

PR #401, opened 2024-05-01
feat(generate): Add generate command and initial component generation

### LINKED ISSUE #272: Component-Definition template generation
Given a known standards Catalog, support imperative and declarative generation of component-definition files in a reproducible manner. 

Questions:
- What information do we need at a minimum to support CLI & manifest driven generation?
  - standard, controls, etc
- How can Lula create a reproducible workflow that allows for humans to add data?
  - Consume existing documents and perform in-place updates.
  - Hash fields for changes?

Objectives:
- Given the catalog source file (json,yaml), generate a component-definition containing declared controls.
  - Imperative & declarative

brandtkeller:
## Description

Initial `lula generate` command with `lula generate component` functionality. 

This performs the creation of a new `component-definition` object with a single `component` and a single `control-implementation`.

This fetches a catalog from some source location - and performs mapping of controls from the catalog to `implemented-requirements`

If an existing component-definition is present - this will perform a merge of two component-definitions such that we look to retain as much of the original content as possible - only overwriting fields solely owned by automation (currently `remarks`).

This merge functionality was written to support future import and merge activities that are planned.

As with all of the automation we have planned for generation - this is a best effort mapping until we can either get more evidence towards best practices or use cases to help guide the process. 

### Assumptions

The logic for populating the remarks is based primarily on the 800-53 rev5 structure of how `parts` is implemented. the `--remarks` flag accepts an array of potential parts to use in populating the remarks and is hopefully extensible enough to enable other catalogs to be used in this process. 

This logic does not include referencing a local catalog source currently. This could be easily supported but ideally catalogs are owned and maintained by a standards or benchmark authority. 

## Related Issue

Closes #272 

## Type of change

## Checklist before merging


src/cmd/generate/generate.go
```diff
@@ -0,0 +1,228 @@
...
+// This will include some prerequisites such as component-definitions - but will ultimately lead to generation
+// of the SSP / SAP / POAM / and maybe SAR in a maintainable way.
+var generateCmd = &cobra.Command{
+	Use:     "generate",
+	Hidden:  false, // Hidden for now until fully implemented
+	Aliases: []string{"g", "gen"},
+	Short:   "Generate a specified compliance artifact template",
+}
+
+var componentHelp = `
```
meganwolf0, line 45: should information about `remarks` and `component` be added here as well?

docs/oscal-generation.md
```diff
@@ -0,0 +1,34 @@
...
+There are optional flags that can be added to the command to generate a component definition:
+- The remarks you wish to map from the control parts to the requirement remarks`--remarks`; `assessment-objective,statement,guidance`
+- The title of the component `--component`; `Software Title`
+- The output file of the component `-o` or `--output`; `oscal-component.yaml`
+
+### Existing Data
+
+The ability to retain data that is put into OSCAL artifacts is of utmost importance to this generation process and also a large feature of continued maintenance of these artifacts. Lula supports the ability to merge newly generated component definition templates into existing component definitions automatically. 
+
+By specifying `--output` or `-o` and providing an existing file - Lula will perform a merge operation that only overwrites specific fields owned by automation.
```
meganwolf0, line 25: It appears that even when I don't provide a `-o` the assessment file is still merged (I assume because then the default file `oscal-component.yaml` exists). Unsure if this should be the behavior or not.

src/pkg/common/oscal/component.go
```diff
@@ -1,29 +1,374 @@
...
+// Creates a component-definition from a catalog and identified (or all) controls. Allows for specification of what the content of the remarks section should contain.
+func ComponentFromCatalog(source string, catalog oscalTypes_1_1_2.Catalog, componentTitle string, targetControls []string, targetRemarks []string) (componentDefinition oscalTypes_1_1_2.ComponentDefinition, err error) {
+	// store all of the implemented requirements
+	implmentedRequirements := make([]oscalTypes_1_1_2.ImplementedRequirementControlImplementation, 0)
+
+	// How will we identify a control based on the array of controls passed?
+	controlMap := make(map[string][]string)
+	for _, control := range targetControls {
+		id := strings.Split(control, "-")
+		controlMap[id[0]] = append(controlMap[id[0]], id[1])
```
meganwolf0: Tested with passing a bogus requirement in the list `-r ac-1,ac-3,ac-3.2,ac4` and this panicked with no information about which requirement was messed up - unsure if it's a huge issue but I wonder if someone passes in a big list of requirements and one has a typo if they'll be able to figure out the problem when the routine breaks down here

src/pkg/common/oscal/component.go
```diff
@@ -1,29 +1,374 @@
...
+		// Is this a group of controls that we are targeting
+		if controlArray, ok := controlMap[group.ID]; ok {
+			if group.Controls == nil {
+				return componentDefinition, fmt.Errorf("group %s has no controls", group.ID)
+			}
+			for _, control := range *group.Controls {
+				id := strings.Split(control.ID, "-")
+				// Check if the control is the primary control
+				if contains(controlArray, id[1]) {
+					newRequirement, err := ControlToImplementedRequirement(control, targetRemarks)
```
meganwolf0: Also tried with `-r ac-1,ac-3,ac-3.2,ac-4.100` where 4.100 is definitely not a control. It passed and gave me no information that the requested control was missing, it might be good to have some debug print statement on a control not being found so you could re-run and add that control in the case of a typo

src/pkg/common/oscal/component.go
```diff
@@ -1,29 +1,374 @@
...
+					ID: param.ID,
+					Select: &selection{
+						HowMany: sel.HowMany,
+						Choice:  *sel.Choice,
+					},
+				}
+			}
+		}
+	} else {
+		message.Debugf("No parameters found for %s", control.ID)
```
meganwolf0: Didn't dig too deep - but when I run the example with `-l debug` this is showing that `No parameters found for ac-3`. But when I look at the output file I'm seeing everything I'd expect to see under ac-3 so mostly just looking for clarity on what this means in that case

src/cmd/generate/generate.go
```diff
@@ -0,0 +1,228 @@
...
+
+		yamlEncoder := yaml.NewEncoder(&b)
+		yamlEncoder.SetIndent(2)
+		yamlEncoder.Encode(component)
+
+		message.Infof("Writing Component Definition to: %s", fileName)
+
+		err = os.WriteFile(fileName, b.Bytes(), 0644)
+		if err != nil {
+			message.Fatalf(fmt.Errorf("error writing Component Definition to: %s", fileName), "error writing Security Assessment Results to: %s", fileName)
```
meganwolf0:
```suggestion
			message.Fatalf(fmt.Errorf("error writing Component Definition to: %s", fileName), "error writing Component Definition to: %s", fileName)
```

docs/oscal-generation.md
brandtkeller: I agree - there is a default here. I believe we need a common function for doing this and I'll capture the intent in #402

src/pkg/common/oscal/component.go
brandtkeller: Updated to no longer fail on malformed controls as well as logging when a control was not found in the catalog with `-l debug`

src/pkg/common/oscal/component.go
brandtkeller: Handled to no-longer panic

src/pkg/common/oscal/component.go
brandtkeller: This is a hard one to describe. Updated the message to reflect that `params` are an optional array field for a control. Essentially storing parameters that will be used to template items in "parts". 

Very much specific to OSCAL.

src/cmd/generate/generate.go
brandtkeller: Updated - good catch

meganwolf0 (approved):

mike-winberry (approved):

PR #407, opened 2024-05-05
feat(oscal): merge on write consolidation

### LINKED ISSUE #402: File merge on write capabilities
Lula will be operating on all of the OSCAL models in some way as context is used across models to build more provenant workflows. This could mean many artifacts that are required for validation and evaluation as well as generation or evidence being required for even a single repository. 

Instead of requiring a file per model - the current types already support a combined definition for all models. When lula performs any write operation to a file - it should read any existing file and merge it into the appropriate model before writing the result. Meaning that models could be co-located in the same file. 

- **Given** a file exists containing some OSCAL data
- **When** any command is executed that writes to a file (generation, validation, etc)
- **Then** Lula supports retaining the existing data and performing a merge as appropriate. 

Keep all files as distinct artifacts where each model is separate of one another.

### Additional context
This should cut down requirements for 1 -> many OSCAL files for every "package"

brandtkeller:
## Description
Common write functionality that should provide consolidation for writing any OSCAL objects to a file. 

## Discussion
Open to feedback or discussion here. Having the ability to aggregate many oscal models into a single artifact will allow for greater portability and less bloat across the OSCAL lifecycle. Something like a production environment would have used most of the models - requiring one file for each may not be optimal. 

Yes - this has the potential to create large files that are not visually appealing. This is more of an opt-in process based on the output file specified. I'd also argue that machine-readable files are just that - machine-readable - and it will be Lulas job to make the underlying data accessible.

## Related Issue

Closes #402 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/validate/validate.go
+++ b/src/cmd/validate/validate.go
@@ -19,8 +19,8 @@ import (
 )
 
 type flags struct {
-	AssessmentFile string // -a --assessment-file
-	InputFile      string // -f --input-file
+	OutputFile string // -o --output-file
+	InputFile  string // -f --input-file
 }
 
 var opts = &flags{}
@@ -54,10 +54,14 @@ var validateCmd = &cobra.Command{
 			message.Fatalf(err, "Generate error")
 		}
 
-		// Write report(s) to file
-		err = WriteReport(report, opts.AssessmentFile)
-		if err != nil {
-			message.Fatalf(err, "Write error")
+		var model = oscalTypes_1_1_2.OscalModels{
+			AssessmentResults: report,
+		}
+
+		// Write the component definition to file
+		err = common.WriteFile(opts.OutputFile, &model)
+		if err == nil {
+			message.Fatalf(fmt.Errorf("error writing component to file"), "error writing component to file")
 		}
 	},
 }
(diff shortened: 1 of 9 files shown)


src/cmd/validate/validate.go
```diff
@@ -54,18 +54,22 @@ var validateCmd = &cobra.Command{
...
-		err = WriteReport(report, opts.AssessmentFile)
-		if err != nil {
-			message.Fatalf(err, "Write error")
+		var model = oscalTypes_1_1_2.OscalModels{
+			AssessmentResults: report,
+		}
+
+		// Write the component definition to file
+		err = common.WriteFile(opts.OutputFile, &model)
+		if err == nil {
```
meganwolf0:
```suggestion
		if err != nil {
```

src/cmd/validate/validate.go
```diff
@@ -19,8 +19,8 @@ import (
 )
 
 type flags struct {
-	AssessmentFile string // -a --assessment-file
-	InputFile      string // -f --input-file
+	OutputFile string // -o --output-file
```
meganwolf0, line 22: The `validateHelp` still references -a, probably should change to -o

mike-winberry (commented):
@meganwolf0 ready for re-review. Made some changes to the `WriteFile` method to:
- handle directories w/ the default filepath name
- return errors instead of `fatal` with a new error for the debug
- updated references to handle the error with the returned error for coherent tracing in the logs.   
___
| also addressed your comments

meganwolf0 (approved):

Issue #397, opened 2024-04-29
Decompose OSCAL and Validations


brandtkeller:
### Is your feature request related to a problem? Please describe.
An emerging theory around the development of validations is how to best capture continual development work. 

One workflow is to decompose a composed OSCAL file to allow for quick feedback loops of updating a given validation and performing /developing tests. 

### Describe the solution you'd like

- **Given** a composed component-definition
- **When** `tools decompose` is executed
- **Then** the file is deconstructed to place validations in a local directory for further development

### Describe alternatives you've considered
Manual processes but much the same

### Additional context
When decomposing - we may not want to put `UUID` in the validation metadata. If that UUID is supposed to change when modified then it may be something we automate. Review OSCAL [identifier use](https://pages.nist.gov/OSCAL/resources/concepts/identifier-use/)

When decomposing -  we should look into removing extra whitespace to ensure readable formatting.

PR #410, opened 2024-05-07
refactor!: #388 update types to use pointers

### LINKED ISSUE #388: [types]: Refactor to use pointer types for non-primitives
- [ ] Refactor lula types to use pointers.
- [ ] Refactor methods to use pointers
- [ ] refactor to do inplace on pointers where necessary

### Additional context
- this will be breaking

### LINKED ISSUE #403: refactor(component): NewOscalComponentDefinition should be refactored
`component.NewOscalComponentDefinition` should be refactored to take either the bytes or the path only. `sigs.k8s.io/yaml` unmarshals `json` as well as `yaml` making the `if/else` extraneous. The method should be refactored to:   

1. only care about the bytes provided 
  - allow unmarshal to fail if not the correct format and return error. 
2. only care about the filepath
  - read the file
  - optional: return the bytes in addition to the `ComponentDefinition`

https://github.com/defenseunicorns/lula/blob/918299ad397363e0d3580cd15b92ddf09044ce05/src/pkg/common/oscal/component.go#L30

mike-winberry:
## Description
- refactor!(common): #388 update common/types (and all references) to use pointers

...

## Related Issue
#388 
#403 

## Type of change

- [x] Breaking refactor

## Checklist before merging


meganwolf0 (commented):
A couple things I noticed:
- When I tried composing some local files I have, I noticed sometimes the rego would be just a giant string (which is real hard to read/edit) and other times it was actually formatted properly. I had both seemingly structured the same by the look of things in the editor, but I assume there's some hidden characters there that's messing with the input/output. Wondering if that's something we could fix? (see `pepr_validations` branch in compliance-artifacts for the pepr.yaml I composed.) I suppose it's down to if we intend people to edit these after they've been composed 
- During this exercise I also noticed that some empty fields in the validation.yaml were being rendered as well (for the respective 'spec' chosen, e.g., Kubernetes-spec shows empty `field` and `wait` fields). Wondering if we should pointer-ize those too? (also possibly add some omitempty's)

mike-winberry:
> * When I tried composing some local files I have, I noticed sometimes the rego would be just a giant string (which is real hard to read/edit) and other times it was actually formatted properly. I had both seemingly structured the same by the look of things in the editor, but I assume there's some hidden characters there that's messing with the input/output. Wondering if that's something we could fix? (see `pepr_validations` branch in compliance-artifacts for the pepr.yaml I composed.) I suppose it's down to if we intend people to edit these after they've been composed


-  this has to do with bad formatting in the yaml, usually spaces in the front of the string, if it is formatted/linted yaml it shouldn't do this
- we had this issue before on the `dev validate` command, and I added something to go-oscal to strip the additional whitespace. Ill look and see if I can do the same when composing 
- added a note to #397 to validate the formatting (strip whitespace) prior to outputting the validations.

mike-winberry:
> * During this exercise I also noticed that some empty fields in the validation.yaml were being rendered as well (for the respective 'spec' chosen, e.g., Kubernetes-spec shows empty `field` and `wait` fields). Wondering if we should pointer-ize those too? (also possibly add some omitempty's)

- This should be fixed now, just need validating

meganwolf0:
> > * During this exercise I also noticed that some empty fields in the validation.yaml were being rendered as well (for the respective 'spec' chosen, e.g., Kubernetes-spec shows empty `field` and `wait` fields). Wondering if we should pointer-ize those too? (also possibly add some omitempty's)
> 
> * This should be fixed now, just need validating

I see the omitemptys for some of the fields, wondering if this was also supposed to address the other empty fields (e.g., `wait`, `field`)? If not no worries, idk if there's a consensus for how those should be handled, if they should show up empty or not. Just checking if something else was supposed to be pushed

mike-winberry:
> > > * During this exercise I also noticed that some empty fields in the validation.yaml were being rendered as well (for the respective 'spec' chosen, e.g., Kubernetes-spec shows empty `field` and `wait` fields). Wondering if we should pointer-ize those too? (also possibly add some omitempty's)
> > 
> > 
> > 
> > * This should be fixed now, just need validating
> 
> I see the omitemptys for some of the fields, wondering if this was also supposed to address the other empty fields (e.g., `wait`, `field`)? If not no worries, idk if there's a consensus for how those should be handled, if they should show up empty or not. Just checking if something else was supposed to be pushed

- Gotcha! Ill fix that, I totally missed the types files in the domains and providers folders...

mike-winberry:
@defenseunicorns/lula-dev should https://github.com/defenseunicorns/lula/blob/13d384a8ac47d9517502ef6cb6d42f171c5b39ae/src/pkg/providers/opa/types.go#L14

Should the providers be pointers as well? the parents have them as pointers but specifically 
```golang
type KyvernoProvider struct {
	// Context is the context that the Kyverno policy is being evaluated in
	Context context.Context `json:"context" yaml:"context"`

	// Spec is the specification of the Kyverno policy
	Spec KyvernoSpec `json:"spec" yaml:"spec"`
}
```
and for opa

meganwolf0:
> Should the providers be pointers as well? the parents have them as pointers but specifically

I assume if we did this with providers, we'd also want to with the domains? I don't really have a strong opinion, maybe would lean toward leaving them as is, since we aren't modifying those elements and they shouldn't ever be nil. Although if we want to change the convention to be fully pointers then maybe yeah?

mike-winberry:
> > Should the providers be pointers as well? the parents have them as pointers but specifically
> 
> I assume if we did this with providers, we'd also want to with the domains? I don't really have a strong opinion, maybe would lean toward leaving them as is, since we aren't modifying those elements and they shouldn't ever be nil. Although if we want to change the convention to be fully pointers then maybe yeah?

Yeah, I think I am just going to do them for consistency, I think that is probably better than not.

meganwolf0:
Did another round of testing this morning and noticed that everything fails with pointer exceptions when we run in the case where resources are not found (I didn't test additional failure conditions) - this wasn't really an issue before but recently we added logic such that any validation failure would be handled gracefully in the form of a remark output in the oscal instead of a panic. I'm guessing really we just need more tests to account for bad validations, missing providers or domains, and empty resources so we can incorporate graceful failure into testing. Unsure if we want to try and wrap those tests into this issue or fix the problems in this PR then incorporate those tests in a separate one.

mike-winberry:
> Did another round of testing this morning and noticed that everything fails with pointer exceptions when we run in the case where resources are not found (I didn't test additional failure conditions) - this wasn't really an issue before but recently we added logic such that any validation failure would be handled gracefully in the form of a remark output in the oscal instead of a panic. I'm guessing really we just need more tests to account for bad validations, missing providers or domains, and empty resources so we can incorporate graceful failure into testing. Unsure if we want to try and wrap those tests into this issue or fix the problems in this PR then incorporate those tests in a separate one.

- Would you mind adding those test cases and pushing, even if they fail, that will help me bolster the tests and find any issues.

meganwolf0 (dismissed):

meganwolf0 (approved):

brandtkeller (approved):

PR #415, opened 2024-05-14
feat(validate)!: #408 create resources in kubernetes domain

### LINKED ISSUE #408: Create a resource in kubernetes domain
Looking at the pepr (or kyverno, gatekeeper, etc) validations, it would be useful to create a dummy resource to test policy enforcement then measure the outcome (ideally denied if some policy you're testing is implemented correctly). Wondering if this feature might have other use cases - thinking along the lines of probing networking and/or making api requests. 

This might also have some baggage associated - because you're not really fully providing "direct proof" via explicit manifest data, you're doing like a proof by non-occurrence - e.g., if a resource is blocked, then the policy is working. That's maybe something that could be interesting to capture as different Lula workflows or reporting?

- **Given** a policy is enacted in the cluster
- **When** a Lula validation wants to validate the policy
- **Then** a dummy resource could be created

None

### Additional context
None

meganwolf0:
## Description

Adds an additional parameter in the Kubernetes domain spec to "create" resources. Purpose being to test admission controllers, networking, and possibly other constraints in a k8s cluster that would otherwise be challenging/impossible to evaluate simply with resource manifests.

## Related Issue

Relates to #408 

## Type of change

## Checklist before merging


src/pkg/domains/kubernetes/create.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return collections, nil
+}
+
+// CreateResourceFromManifest() creates the resource from the manifest string
+func CreateFromManifest(ctx context.Context, client klient.Client, resourceBytes []byte) ([]map[string]interface{}, error) {
+	resources := make([]map[string]interface{}, 0)
+
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(resourceBytes), 4096)
+	for {
```
mike-winberry: This would be a good helper method to be or be a part of how we can standardize yaml parsing.

meganwolf0: I couldn't really figure out how to pull this out in widely reusable function, so I created a few specific ones for the data types that might be showing up in arrays. Let me know if you see a better path forward though.

__debug_bin1450439560
mike-winberry, line 1: straggle debug file?

mike-winberry (changes requested):

mike-winberry (dismissed):

src/cmd/validate/validate.go
```diff
@@ -153,136 +155,43 @@ func ValidateOnCompDef(compDef *oscalTypes_1_1_2.ComponentDefinition) (map[strin
...
-				spinner.Success()
+	// Create requirement store for all implemented requirements
+	requirementStore := requirementstore.NewRequirementStore(compDef)
+	message.Title("\n🔍 Collecting Requirements and Validations", "")
+	requirementStore.ResolveLulaValidations(validationStore)
+	reqtStats := requirementStore.GetStats(validationStore)
+	message.Infof("Found %d Implemented Requirements", reqtStats.TotalRequirements)
+	message.Infof("Found %d runnable Lula Validations", reqtStats.TotalValidations)
+
+	// Check if validations perform execution actions
```
brandtkeller, line 178: I'd like to discuss the default workflow as a team and consider UX more.

It looks like Lula will check for executable validations -> If present will then prompt the user to allow executing -> if yes then validate -> if no then cancel everything.

Wanting to step back and consider UX as I see this problem on zarf all the time. "I want to do x non-interactively" and it's supported but through a combination of flags rather than the default. 

One answer may be extending the confirmation such that a user can select no executable but still perform all non-executable (via interactive options).

src/pkg/domains/kubernetes/create.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil, err
+	}
+	for _, obj := range objArray {
+		resource, err := createResource(ctx, client, &obj)
+		if err == nil {
+			resources = append(resources, resource.Object)
+		}
+	}
+
+	// decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(resourceBytes), 4096)
```
brandtkeller: Clear to remove following commented code?

src/pkg/common/oscal/component.go
```diff
@@ -401,6 +407,32 @@ func BackMatterToMap(backMatter oscalTypes_1_1_2.BackMatter) (resourceMap map[st
...
+
+	if componentDefinition.Components == nil {
+		return requirementMap
+	}
+
+	for _, component := range *componentDefinition.Components {
+		if component.ControlImplementations != nil {
+			for _, controlImplementation := range *component.ControlImplementations {
+				for _, requirement := range controlImplementation.ImplementedRequirements {
+					// TODO: should this be controlID (i.e., possibly combining multiple instances of the same control?)
```
brandtkeller: Want to make sure this comment was resolved? Looks like you identified correctly that `controlId` would not make a good unique identifier.

src/cmd/validate/validate.go
```diff
@@ -153,136 +155,43 @@ func ValidateOnCompDef(compDef *oscalTypes_1_1_2.ComponentDefinition) (map[strin
...
-
-				// Only add related observations if there are any present (preventing an empty array in the field)
-				if len(relatedObservations) > 0 {
-					finding.RelatedObservations = &relatedObservations
-				}
-
-				findings[implementedRequirement.ControlId] = finding
-				observations = append(observations, tempObservations...)
-				spinner.Success()
+	// Create requirement store for all implemented requirements
```
brandtkeller, line 170: Should this functionality move to a library in some way?

src/pkg/common/requirement-store/requirement-store.go
```diff
@@ -0,0 +1,141 @@
...
+
+// ResolveLulaValidations resolves the linked Lula validations with the requirements and populates the ValidationStore.validationMap
+func (r *RequirementStore) ResolveLulaValidations(validationStore *validationstore.ValidationStore) {
+	// get all Lula validations linked to the requirement
+	var lulaValidation *types.LulaValidation
+	for _, requirement := range r.requirementMap {
+		if requirement.ImplementedRequirement.Links != nil {
+			for _, link := range *requirement.ImplementedRequirement.Links {
+				if common.IsLulaLink(link) {
+					id := common.TrimIdPrefix(link.Href)
```
brandtkeller: Is this function redundant if `validation.GetLulaValidation(id)` calls it again? (or reverse)

brandtkeller (changes requested):
This is generating a lot of great discussion and potential. Few items I'd like to address as a team. May have even found a few unrelated bugs that need fixing.

src/cmd/validate/validate.go
meganwolf0: Yeah I think actually the default case is to continue executing the non-executable ones only - if you saw a different behavior then can you point me to the test file you tried with? But it might be good to make it like a few part option -> continue with non-executable or just stop. And I think yes, we probably do need to add a non-interactive flag option to at least allow this to run in a pipeline

src/cmd/validate/validate.go
meganwolf0: Possibly... I tried to break out a lot of this to requirement-store to try and libratize(?) but there also may be a better/more encapsulated way to structure all this. The part where the comment is - on the messages to the user - feels more cli-esque so wasn't really thinking of that as a library, but perhaps some of the requirement-store functions there maybe could/should be organized differently. Idk, open to alternate thoughts on how this could look!

src/pkg/message/interactive.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func PromptForConfirmation(spinner *Spinner) bool {
+	// Prompt the user to confirm the action
+	if spinner != nil {
+		spinnerText := spinner.Pause()
+		defer spinner.Updatef(fmt.Sprintf("%s\n", spinnerText))
+	}
+
+	confirmation := pterm.DefaultInteractiveConfirm.WithDefaultText("Do you want to continue?")
```
brandtkeller:
```suggestion
	confirmation := pterm.DefaultInteractiveConfirm.WithDefaultText("Do you want to run executable validations?")
```

OR similar -> current process makes me think that inputting "no" would cancel all runtime

src/cmd/validate/validate.go
brandtkeller: Resolving as I think this moves us closer. I imagine we need a library function in the future that exposes validating a component-definition in some way. From there we can resolve how to handle logging and user interaction.

src/pkg/domains/kubernetes/create.go
meganwolf0: removed

src/pkg/common/oscal/component.go
meganwolf0: removed

src/pkg/common/requirement-store/requirement-store.go
meganwolf0: yep I agree it's redundant, removed

src/cmd/validate/validate.go
brandtkeller: Correct me if i am wrong here -  we cannot run validation on OSCAL containing both executable and non-executable validations in a non-interactive format when we only want the non-executable validations? (IE CI/CD)

meganwolf0: yep that's right, I think we'd need to add another flag for like non-interactive or something such that if non-interactive was flagged and confirm-execution was not, then all non-executable validations would run without prompt. That path doesn't currently exist. I held off on that because I wasn't sure if that was a path we wanted to support, but I think it should be a pretty benign addition as I already have the variable in the validation file it just would need to turn into a cmd flag

brandtkeller: I see it as a necessary path - executable validations will be important on building provenance - but we'll also want defined CI workflows that only run the non-executable workflows.

src/pkg/common/validation-store/validation-store.go
```diff
@@ -71,3 +85,98 @@ func (v *ValidationStore) GetLulaValidation(id string) (validation *types.LulaVa
...
+	}
+	return false, "No validation is executable"
+}
+
+// RunValidations runs the validations in the store
+func (v *ValidationStore) RunValidations(confirmExecution bool) []oscalTypes_1_1_2.Observation {
+	observations := make([]oscalTypes_1_1_2.Observation, 0, len(v.validationMap))
+	for k, val := range v.validationMap {
+		completedText := "evaluated"
+		spinnerMessage := fmt.Sprintf("Running validation %s", k)
```
brandtkeller, line 109: I believe we'll want to be more careful about our embedding of the message package in the future. Not a change required now but someone attempting to use these functions outside Lula may come knocking sooner-than-later. 

(Resolve once read)

src/pkg/common/validation-store/validation-store.go
```diff
@@ -71,3 +85,98 @@ func (v *ValidationStore) GetLulaValidation(id string) (validation *types.LulaVa
...
+		spinner := message.NewProgressSpinner(spinnerMessage)
+		defer spinner.Stop()
+		err := val.Validate(types.ExecutionAllowed(confirmExecution))
+		if err != nil {
+			message.Debugf("Error running validation %s: %v", k, err)
```
brandtkeller, line 114: Do we need some documentation that describes the output of this behavior and how it impacts the OSCAL produced? 

Current implementation will mean that any `implemented-requirement` that has both non-executable and executable validations AND is validated without the executable validations will result in the `finding` being `not-satisfied`.

I'd offer another scenario to discuss which is when we do not run executable validations - they are ignored and do not impact the findings.

IE I expect executable validations to be run against dev/test/staging environments BUT when promoted to production I still want to evaluate my assessment results without executable validations. 

we don't have any real data to support what people would optimally want here - but I err on the side of caution.

meganwolf0: Yeah, I think this is probably feeding into a larger conversation about when to mark something satisfied / not-satisfied. If I think back to the whole placeholder discussion - I see the non-execution of these policies as turning into default, not-satisfied placeholders such that a user would want to manually evaluate the specific aspect for satisfaction. And by keeping it not-satisfied we're basically telling the user/reviewer there's more work to be done. I suppose I'm being more conservative in just making things not-satisfied but would be interested in the best approach from a larger compliance and reporting standpoint.

I will include additional documentation on this - all the validations will default to `not-satisfied` if they can't be evaluated so a) yes good to outline that and b) probably good to have an internal conversation on if we want to modify that decision in general. Since we only have a binary decision, I would think we shouldn't default to `satisfied` but instead add a third option such as `not-evaluated` or something if we want the distinction.

meganwolf0: added `docs/assessment-results.md` to try and address. I realized we don't have any documentation on Lula <> OSCAL artifacts (short of the ADRs), maybe we need to add an issue for that if appropriate?

brandtkeller: I can appreciate being conservative here. I'm trying to project "How would this work in Production" and I may be very wrong. 

I hold an _assumption_ that we want to report on all non-executable validations while still understanding and reporting controls of-which all other validations pass. I am concerned we'll dilute the results in efforts of being conservative to the point where everything will still report `not-satisfied` given some extenuating circumstance. 

This is a tough one...

CloudBeard: Would the reason field of other make better sense for the not-evaluated ?

https://pages.nist.gov/OSCAL-Reference/models/v1.1.2/assessment-results/json-reference/#/assessment-results/results/findings/target/status/reason

I think assessors will look at it more cut and dry but I do see the need to call out that they didn't pass because there wasn't anything to pass/test against. Both internally so you know the difference in it didn't pass becuase we don't have a test and the opposite that. That would give a way to pull relative the needed info out for the why but still show its not met?

mike-winberry: After reading through the oscal docs, I agree with @CloudBeard that we could establish our own "reason" or use the "other" as one of the "defaults" OSCAL recommends, but it can be anything, I would also probably lean toward the not satisfied, but theres in interesting use case for when it is all satisfied in the dev env, and how those two link up.

mike-winberry: I also think we are limited to `satisfied` `not-satisfied` based on [state](https://pages.nist.gov/OSCAL-Reference/models/v1.1.2/assessment-results/json-reference/#/assessment-results/results/findings/target/status/state)
>The value must be one of the following:
> - satisfied: The objective has been completely satisfied.
> - not-satisfied: The objective has not been completely satisfied, but may be partially satisfied.

brandtkeller: Created #478 to capture the intent of this larger discussion. Resolving this thread.

brandtkeller (dismissed):

brandtkeller (approved):

mike-winberry (approved):

Issue #431, opened 2024-05-21
E2E ADR + Docs


brandtkeller:
### Describe what should be investigated or refactored

The E2E vision for Lula operation is ambiguous as far as intent and vision. An ADR should set expectations as far as intent and design decisions and docs should be authored to break the concepts down into discrete chunks for end user consumption.

PR #432, opened 2024-05-21
chore(adr): add ADR for E2E intent and design

brandtkeller:
## Description

ADR for Lula E2E OSCAL design decisions and expectations.

To follow this ADR:
- Doc for the lifecycle of a software artifact
- Doc for the lifecycle of a target environment
- Future Docs for the generation of the remaining models and the inclusion of them in the lifecycles

## Related Issue

Relates to #431 

## Type of change

## Checklist before merging


adr/0008-end-to-end-oscal.md
```diff
@@ -0,0 +1,97 @@
...
+
+Lula will operate against the intent of a Catalog baseline for the purpose of the intended context for which GRC information is being authored. The context will vary dependent on where in the lifecycle the information is being produced and consumed in the following pattern:
+- A `component definition` will perform the mapping of one-to-many standards/benchmarks/policies to one-to-many system components
+  - Where possible, a single `component definition` should focus on a single system component
+- A `system security plan` will aggregate the collection of one-to-many `component definitions` into a single system definition
+- An `assessment plan` will aggregate one-to-many `component-definitions` into a plan with depth of control satisfaction based upon many layers of implementation for any given control
```
CloudBeard, line 31: Should it be the assessment plan will aggregate one-to-one `system-security-plan` ?

Thinking the assessment plan links to the ssp which links to the components. It technically assess the admin controls too. 

https://pages.nist.gov/OSCAL-Reference/models/v1.1.2/assessment-plan/json-reference/#/assessment-plan/import-ssp

brandtkeller: Good question - it may vey well be. 

If the data flow is: `component -> assessment plan -> system security plan` then yes that may be the case?

But I see a lot of overlap between components and SSP that makes me wonder about this workflow. 

I'm going to look at this ADR again - I don't want to focus too much on the solutioning (hence why there was mention of many possible workflows) -  But I think I need to highlight another point - which is that we might use components to generate an AP & SSP - but when the generation of say the AP detects that an SSP exists - it can establish context sharing. So as you implement each model - you may have cross model connections being made.

CloudBeard: _"we might use components to generate an AP & SSP - but when the generation of say the AP detects that an SSP exists - it can establish context sharing. So as you implement each model - you may have cross model connections being made."_

^ This, 100% agree. I also agree there are several possibilities for our workflows.

I think OSCAL Data Flow viewpoint is 

```mermaid
graph TD;
    A[Component Definition]
    B[Assessment Plan]
    C[System Security Plan]
    D[Assessment Results]
    E[POAM]

    A --> C --> B --> D --> E
 
```
This may show it a little better too. https://pages.nist.gov/OSCAL/resources/concepts/layer/assessment/assessment-results/#key-concepts

CloudBeard (approved):
Big fan of the mermaid flow chart.

mike-winberry (approved):

meganwolf0 (commented):
Looks good to me! I think the one thing thing missing for me was discussion of `evaluate` and how that fits into the end-to-end

brandtkeller:
> Looks good to me! I think the one thing thing missing for me was discussion of `evaluate` and how that fits into the end-to-end

That's a good mention. `Evaluate` does not produce any artifacts (yet), nor is it comprised of any specific artifact itself (IE like a validation). 

Focusing on artifact strategy here - but am open to including process if we believe it is necessary and would warrant a later ADR?

meganwolf0:
> That's a good mention. `Evaluate` does not produce any artifacts (yet), nor is it comprised of any specific artifact itself (IE like a validation).

I was thinking that since it evaluates two assessment-results, it is kind of reliant on OSCAL data and would potentially fit within an E2E picture? But I realize that it's not actually producing any kind of recognized OSCAL data so maybe doesn't actually make sense here. I was more just thinking about enumerating the use case for evaluate being relevant in the full end to end.

I'm good to table the question on evaluate if it's out of scope for this

brandtkeller:
Closing in order to bring properly up-to-date

PR #457, opened 2024-05-31
fix(evaluate): establish threshold for assessment results result

### LINKED ISSUE #439: Identifier for evaluation threshold
`lula evaluate` is designed to be used to compare two assessments (a threshold assessment and a new assessment) and provide a pass/fail exit output and exit code as a result of being more/equally compliant (pass) or less compliant (fail). 

Current this operates on the assumption of two separate assessment files being compared OR the two latest results in a single assessment file.

As Lula looks to make artifact consolidation into a single-file a first class process - we need the ability to store many assessments-results into a single object AND establish a way to identify which result is the threshold.

Example:
I may have an assessment result with a single finding (currently the threshold) - I perform another assessment (validate) which is less compliant. I now have two results - if I validate again, current logic would use the 2nd result (which was failing) as my new threshold. Instead I still want to use the original threshold to ensure I maintain a specific level of compliance until the assessment has become more compliant and the threshold needs to be updated

- **Given** assessment results are being written to a single file
- **When** evaluate is executed
- **Then** the threshold should be a specific result with some identifier

### Additional context
Might be room here for `props` to be introduced

brandtkeller:
## Description

Establish an identifier for which assessment results result is the threshold to be used for evaluation.

Identifier:
```yaml
props:
  - name: threshold
    ns: https://docs.lula.dev/ns
    value: "true/false"
```

This provides a baseline for how we select results for evaluation. 

Refactored the evaluation logic to use library functionality without any embedded logging logic for use by the command logic. 

Identified constraints for the changes:
- This fixes evaluation behavior to create more structure in identifying a threshold given that there may be many `assessment results` results to review. 
- We do not have any data to support whether to use the latest threshold or find a way to reconcile the most valuable threshold as of yet. 

## Related Issue

Closes #439 

## Type of change

## Checklist before merging


CloudBeard (dismissed):
Nothing stands out to me.

mike-winberry (approved):

CloudBeard (approved):

meganwolf0 (approved):

PR #481, opened 2024-06-13
feat: add looping for lint

### LINKED ISSUE #480: feat: add the functionality to loop multiple oscal files in lint
Currently the `lula tools lint` only supports 1 input. As we begin adding Lula into CI I can see the need to lint various oscal files within the same PR.

This functionality will also simplify the maru task that does the linting in each repo it is included in.

Allow for Lula to handle `lula tools lint -f oscal1.yaml, oscal2.yaml...` by listing all of the pass and failed lintings after the command has run.

### Additional context
This functionality allows for maru tasks to pass in multiple files simplifying the tasks themselves by allowing Lula to handle the looping.

CloudBeard:
## Description

Adds loop for lula lint.

Removed mention of yaml to future proof for xml (sorry for anyone reading that manually) and json.

Changed wording around single files to multiple files.

## Related Issue

Fixes: https://github.com/defenseunicorns/lula/issues/480 

## Type of change

## Checklist before merging


brandtkeller (changes requested):
Great first pass - I have a few modifications we should consider: 

This will fail on the first document that fails linting and exit - optimally the workflow would lint all files and report on which are passing AND failing in some capacity. This prevents the cycle of fixing one file at a time. 

This could be iterating through the loop as implemented - identifying pass/fail with some information and once all files have been evaluated then indicating overall success/failure. 

Looking to prevent scope creep here - I do not believe:
```
validation.WriteValidationResult(validationResp.Result, opts.ResultFile)
```
 will handle appending to a result file (currently intended for a single write operation). Meaning we need to update the go-oscal function or develop a stop-gap in the interim. 
 
 Focus on the core logic - will evaluate go-oscal

Open to feedback

src/cmd/tools/lint.go
```diff
@@ -26,32 +26,52 @@ func init() {
...
+
+				message.Infof("Successfully validated %s is valid OSCAL version %s %s\n", inputFile, validationResp.Validator.GetSchemaVersion(), validationResp.Validator.GetModelType())
+				spinner.Success()
 			}
 
 			if opts.ResultFile != "" {
-				validation.WriteValidationResult(validationResp.Result, opts.ResultFile)
+				err := validation.WriteValidationResults(validationResults, opts.ResultFile)
+				if err != nil {
+					message.Fatalf(err, "Failed to write validation results to %s\n", opts.ResultFile)
```
brandtkeller:
```suggestion
					message.Fatalf(err, "Failed to write validation results to %s with error: %s", opts.ResultFile, err.Error())
```

This adds more context around why the error happened -IE if I supplied `-r output.xml` it will tell me that it failed to write AND that the file type can only be yaml or json.

src/cmd/tools/lint.go
```diff
@@ -26,32 +26,52 @@ func init() {
...
 
-			for _, warning := range validationResp.Warnings {
-				message.Warn(warning)
+				validationResp, err := validation.ValidationCommand(inputFile)
+
+
+				if err != nil {
+					message.Warnf("Failed to lint %s: %v\n", inputFile, err)
+					errorsOccurred = true
+					continue
```
brandtkeller: The presence of this `continue` will prevent the `ValidationResult` from being appended below and as such the failed results will not be present in the output file. 

Given there may be a number of errors that could have been returned from the function -> I would: 
- attempt to catch known errors for failed linting (or otherwise halt execution with a fatal log)
- proceed to still log warnings as done below (no modification needed)
- proceed to add all results to the slice as done below (no modification needed)
- and then add a conditional at the end with `validationResp.Result.Valid` logging success - else logging failed linting

brandtkeller (changes requested):

mike-winberry (commented):
Agree with @brandtkeller's changes, nothing else stood out me!

src/cmd/tools/lint.go
```diff
@@ -26,32 +26,54 @@ func init() {
...
+			for _, inputFile := range opts.InputFiles {
+				spinner := message.NewProgressSpinner("Linting %s\n", inputFile)
+				defer spinner.Stop()
 
-			for _, warning := range validationResp.Warnings {
-				message.Warn(warning)
+				validationResp, err := validation.ValidationCommand(inputFile)
+
+				if err != nil {
+					if validatorErr, ok := err.(wrappedValidatorError); ok {
```
brandtkeller: Can you point us to where `wrappedValidatorError` originates?

CloudBeard: From local testing of a wrapper. It worked having a wrapper so I could have assertions on validation.ValidationError to check for the know error types in the validationError.

I may also be going down the wrong path to solve for: attempt to catch known errors for failed linting (or otherwise halt execution with a fatal log)

So in go-oscal validationError.go the wrapper would be 

```go
func (e ValidatorError) Error() string {
	return e.Error
}
```

In lint.go id swap that a bit wording wise to 

```go
if err != nil {
    if validatorErr, ok := err.(validation.ValidatorError); ok {
        message.Fatalf(err, "Validation error occurred while linting %s: %v\n", inputFile, validatorErr)
    } else {
        message.Warnf("Failed to lint %s: %v\n", inputFile, err)
        errorsOccurred = true
    }
}
```

mike-winberry:
- [x] depends on https://github.com/defenseunicorns/go-oscal/issues/266

mike-winberry:
- [x] depends on #492

mike-winberry (approved):

src/cmd/tools/lint.go
```diff
@@ -28,42 +28,55 @@ func init() {
...
+
+				for _, warning := range validationResp.Warnings {
+					message.Warn(warning)
 				}
-				message.Infof("Validation result: %s", string(jsonBytes))
-			}
 
-			if validationResp.JsonSchemaError != nil {
-				message.Fatalf(err, "Failed to lint %s", opts.InputFile)
+				if opts.ResultFile != "" {
```
mike-winberry: Do we want to extract the results into a slice and the use `validation.WriteResult` if `len == 1` else `validation.WriteResults` (added in go-oscal v0.5.0)?

mike-winberry (changes requested):

src/cmd/tools/lint.go
mike-winberry: https://github.com/defenseunicorns/lula/pull/504 fixes ⬆️

mike-winberry (approved):

brandtkeller (approved):
Operating as expected - walked through a few workflows and received intended results.

PR #495, opened 2024-06-20
chore(docs): re-organize docs for docs website consumption

### LINKED ISSUE #490: Docs organization and headings
For integration with the [lula-docs](https://github.com/defenseunicorns/lula-docs) website - the directory structure of `/docs` should have sub-directories for website sections

Add sub-directories for proposed headings

### Additional context
The website repository linked above should become public when it becomes stable.

brandtkeller:
## Description

Attempt to set new docs organization for the lula-docs website. 

## Related Issue

Fixes #490 

## Type of change

## Checklist before merging


docs/cli-commands/generate/component.md
```diff
@@ -1,4 +1,4 @@
-# OSCAL Generation
+# Component
```
brandtkeller, line 1: @bdfinst do we need a `README.md` at the root of each sub-directory?

meganwolf0:
Is there a good example to look at for how the file/folder structure maps to the website UI?

docs/cli-commands/generate/component.md
bdfinst: The current assumption is that README.md or index.md is the entry point for the docs in the subdirectory. 

However, this is only true for the top level of menus. 
So, if we have `docs` with siblings `docs/topic1` and `docs/topic2`, we can make `topic1` and `topic2` top-level menu items by adding one of those files inside. Their child subdirectories would not need (or want) a README or index file.
In that scenario, `docs` would have no markdown in it. Only its subdirectories would.

bdfinst:
> Is there a good example to look at for how the file/folder structure maps to the website UI?

You can see an example here:
https://github.com/defenseunicorns/uds/tree/main/content/en

meganwolf0:
I think this looks good, I have a few thoughts, but I know we don't really have the documentation to support yet so really just jotting down notes somewhere for discussion
- I think we need like a "what is lula/about" section that details more of the use cases and how to use Lula, perhaps a subsection we'd have "building a validation" or something similar. I'm thinking this could link to the other docs in oscal and references
- I think we probably need some more detailed information for the lula validation yaml, which could be an autodoc generated from the schema
- Wondering if we should repurpose the `evaluate` into like a `quickstart` with details on the `validate` command as well.

brandtkeller:
> I think this looks good, I have a few thoughts, but I know we don't really have the documentation to support yet so really just jotting down notes somewhere for discussion
> 
> * I think we need like a "what is lula/about" section that details more of the use cases and how to use Lula, perhaps a subsection we'd have "building a validation" or something similar. I'm thinking this could link to the other docs in oscal and references
> * I think we probably need some more detailed information for the lula validation yaml, which could be an autodoc generated from the schema
> * Wondering if we should repurpose the `evaluate` into like a `quickstart` with details on the `validate` command as well.

I completely agree - I believe we have more than a few efforts for documentation to support going forward. Let's get some structure in-place such that we can wire docs up to the website and iterate.

CloudBeard (approved):

meganwolf0 (approved):

PR #502, opened 2024-06-25
fix(oscal): single model write operations support

### LINKED ISSUE #501: Write single models to remain schema compliant
### Expected result

Only allow single model per file

### Actual Result

allows for multiple models per file

### Severity/Priority
high

brandtkeller:
## Description

Only allow a single model per file write - validate that is the case

## Related Issue

Fixes #501

## Type of change

## Checklist before merging


src/pkg/common/oscal/complete-schema.go
```diff
@@ -112,3 +127,44 @@ func MergeOscalModels(existingModel *oscalTypes_1_1_2.OscalModels, newModel *osc
 
 	return existingModel, err
 }
+
+func GetOscalModel(model *oscalTypes_1_1_2.OscalModels) (modelType string, err error) {
+
+	// Check if one model present and all other nil - is there a better way to do this?
```
mike-winberry, line 133: Performance wise, I think this is probably the best, the only other things I can think of are reflection, map conversion, (un)marshalling all of which may be pointlessly complex and slow rn. Though this could lead to needed maintenance that could be avoided with those options...Looking at WriteOscalModel, it may be possible to re-arrange the order of ops, to save that performance hit, since we will be writing and converting back to bytes, so it would only be one additional marshal to convert to a map and then remove the need for the hardcoded strings (will be in json/yaml form) and may remove the need for the hardcoded checks. May be worth an issue, don't think its worth blocking for though, unless I missed something.

mike-winberry (changes requested):

src/pkg/common/oscal/complete-schema.go
mike-winberry: #503

mike-winberry (approved):

CloudBeard (approved):

src/pkg/common/oscal/complete-schema.go
brandtkeller: agreed! Considered other options but kept coming back to the same complexity for the same problem. Kept this one readable (albeit verbose) for traceability. Appreciate the issue.

PR #524, opened 2024-07-08
chore(docs): initial docs structure/changes for feedback

### LINKED ISSUE #521: Doc clean-up

Docs need cleaned up, going to use this as a blanket issue to propose some documentation organization and content changes.

meganwolf0:
## Description

Revamped docs content and structure

## Related Issue

Fixes #521 

## Type of change

- [x] Other (docs update,)

## Checklist before merging


meganwolf0:
Some notes:

Rendering this from the `lula-docs` repo, some links appear to be broken (particularly on the README). I'm thinking that once the site is actually live, we'll just point our links to the actual site instead of the relative links to our docs folder.

Also, I did a bit of un-nesting/renaming with some of the directories and readmes, mostly because I thought it just looked too clunky in the web interface

CloudBeard (approved):
Changes look good to me! I also think you're right about changing the doc pointers but im not 100%

(I broke your branch when I merged main in, quick fix just added the new words)

brandtkeller (approved):
Love the direction this sets for the docs. some content is forward leaning but I appreciate that it exists. We may need to adapt `lula-docs` for the lula.svg as discussed but that shouldn't be difficult.

PR #532, opened 2024-07-11
fix(validate): allow for optionality among potential standards in a component definition

### LINKED ISSUE #327: Optionality for validation/mapping of Comp-Def control-implementations
## Context

TLDR: Add support for validating/mapping component-definitions with many control-implementations. 

The component-definition exists as a layer below catalogs/profiles such that it captures more finely scoped context about specific components and the controls (implemented-requirements) that a component can potentially satisfy.

Lula adds a flavor of execution around component-definitions and allows for shifting compliance left given the existence of this single artifacts with validations Lula can perform. 

As Lula looks to target validation of assessment-plans (mapped from component-definitions), the potential that Lula would want to support a component-definition with many control-implementations is more likely to occur. This also allows for treating component-definitions as another source of different standards mapping while allowing re-use of validations across standards

## Objectives

- `lula validate` flags for specification of which control-implementation to validate
- Establish default behavior when not specified

### LINKED ISSUE #499: Findings results from validate repetitive
### Environment
Device and OS:  
App version:    
Kubernetes distro:  
Kubernetes version:  
provider:
- [ ] kyverno
- [ ] opa

### Steps to reproduce
1. Aggregate multiple components that specify the same control
2. `lula validate -f <oscal.yaml>`

### Expected result
For each "finding" the status should be printed once:
💡 Findings   
  •  UUID: 14cb0f0d-c8dd-4225-a5fa-bc08b6a5a78d
  •  Status: satisfied
  •  OSCAL artifact written to: assessment.yaml

### Actual Result
If multiple implemented requirements reference the same control ID, the findings are printed multiple times:
💡 Findings   
  •  UUID: 14cb0f0d-c8dd-4225-a5fa-bc08b6a5a78d
  •  Status: satisfied
  •  UUID: 14cb0f0d-c8dd-4225-a5fa-bc08b6a5a78d
  •  Status: satisfied
  •  OSCAL artifact written to: assessment.yaml

### Severity/Priority
Low

### Additional Context
requirementstore.GenerateFindings runs the logger that prints out the result, this should be reorganized such that this is only evaluated once per finding, not per finding evaluation (as the current state)

### LINKED ISSUE #500: Assessment Findings don't account for multiple components evaluating against the same Control ID
### Environment
Device and OS:  
App version:    
Kubernetes distro:  
Kubernetes version:  
provider:
- [ ] kyverno
- [ ] opa

### Steps to reproduce
1. Aggregate multiple components that specify the same control
2. `lula validate -f <oscal.yaml>`

### Expected result
Assessment results model should report on each Finding, or at least document the individual implemented requirement text/information

### Actual Result
Findings takes only one implemented requirement text/data, which omits information that might be relevant/important from other implemented requirements

### Severity/Priority
Medium

### Additional Context

brandtkeller:
## Description

Component definitions are comprised of 1-> components of 1 -> N Control Implementations of 1 -> N Implemented Requirements

The intent of `validate` should support a default behavior of assessing all control implementations (containing common control-id's) and producing a result per targeted source of controls. 

In doing so - Lula can then support validate/evaluate of specific control-implementations via a specified flag instead of needing to execute against non-applicable standards/controls. This will also be an expected behavior for generation of an `assessment-plan` where we will need to know which standard/framework to choose. 

## Core Functionality Updates

- Introduces a `framework` and `target` Lula oscal prop
  - `framework` will be used in the component definition and allows for the combination of multiple standards/control-implementations into a single result
  - `target` is a generated prop that is used by Lula to conduct `evaluate` in a targeted manner vs the filtering and evaluation of all unique "targets"
- Adds a `-t` / `--target` flag to validate and evaluate
- The `assessment-result` is now generated from a slice of results.
- Each unique `control-implementation` identified will create a unique result
  - When the `framework` prop is utilized an additional result will be created for the framework
  - This allows targeting a single standard/framework OR creating results for all standards/frameworks by default

## Constraints

- Proposed changes remove the connection between an implemented-requirement/control-implementation and it's subsequent component (which wasn't thoroughly leveraged before. I'd like to consider implications and determine if adding this back in the future is helpful.
- This update is not entirely backwards compatible. Existing thresholds will need to be updated after determining what the new execution means for impact (if using multiple standards). 

## Local Testing

- Unit tests and E2E tests were updated to fix removed/modified functions
- see `src/test/unit/common/oscal/valid-multi-component.yaml` as a target for performing various permutations of validate/evaluate and noting the amount of results created

## Release Note

TODO: insert release not here for how to adopt these changes in an existing validate/evaluate workflow with an existing threshold. 

## Related Issue

Closes #327 
Closes #499 
Closes #500
## Type of change

## Checklist before merging


docs/ns/framework.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Example
+
+This use of this prop is defined in `control-implementation[_].props` in the following format:
+
+```yaml
+props:
+  - name: framework
+    ns: https://docs.lula.dev/ns
+    value: Critical
```
CloudBeard: For the value of `framework` something like `impact-level-x` may resonate better in the compliance community instead of `Critical`.

Only concern would be relating this to CVE severity.

src/pkg/common/requirement-store/requirement-store.go
```diff
@@ -65,14 +65,15 @@ func (r *RequirementStore) GenerateFindings(validationStore *validationstore.Val
...
+		// A single finding should be "control-id centric"
 		if _, ok := r.findingMap[requirement.ImplementedRequirement.ControlId]; ok {
 			finding = r.findingMap[requirement.ImplementedRequirement.ControlId]
+			finding.Description += fmt.Sprintf("Control Implementation: %s / Implemented Requirement: %s\n %s\n", requirement.ControlImplementation.UUID, requirement.ImplementedRequirement.UUID, requirement.ImplementedRequirement.Description)
 		} else {
 			finding = oscalTypes_1_1_2.Finding{
 				UUID:        uuid.NewUUID(),
-				Title:       fmt.Sprintf("Validation Result - Component:%s / Control Implementation: %s / Control:  %s", requirement.Component.UUID, requirement.ControlImplementation.UUID, requirement.ImplementedRequirement.ControlId),
-				Description: requirement.ImplementedRequirement.Description,
+				Title:       fmt.Sprintf("Validation Result - Control:  %s", requirement.ImplementedRequirement.ControlId),
```
meganwolf0: should there be two spaces between ":" and the control ID? Just looked a little weird to me in the assessment results, wasn't sure if typo or intended

src/pkg/common/requirement-store/requirement-store.go
```diff
@@ -65,14 +65,15 @@ func (r *RequirementStore) GenerateFindings(validationStore *validationstore.Val
...
+				Description: fmt.Sprintf("Control Implementation: %s / Implemented Requirement: %s\n %s\n", requirement.ControlImplementation.UUID, requirement.ImplementedRequirement.UUID, requirement.ImplementedRequirement.Description),
```
meganwolf0: This is a nit, but when an extra space is added after the newline `%s\n %s\n` instead of `%s\n%s\n` it messes up the formatting in the resulting yaml doc so that description blocks won't collapse correctly, which is mildly annoying when looking through the doc and trying to collapse sections. Wondering if we could change to the latter in both this line and above just for readability? 
edited to add - all the indented spaces that get added to description and remarks also messes this up.. I wonder if we should strip those as well to be more yaml compliant?

meganwolf0:
A couple other questions/comments as I'm reviewing:
- Looks like observations aren't all appended to `related-observations` when two different components are calling the same control (perhaps we need a test case for this scenario)
- This one is more personal preference - but the "Findings" section in the validate output is taking up a lot of space in the terminal, especially just with line breaks, would it be possible to condense those to a single line -> Control ID: %s, Finding UUID: %s, Status: %s? Maybe there's a reason they should be that expansive though so happy to hear the counterargument there.
- How does `framework` get added? It appears to be manual right now, is that something that ideally will happen via `lula generate`? 
- When I run `lula validate` against the `valid-multi-component.yaml` I get 4 results - I thought there'd be two for just each of the "frameworks" specified, but seems like the catalog source also gets added as a target... can you ELI5?

CloudBeard:
I noticed the 4 findings but I think it's a good thing. 

You'll get one per framework and one per catalog/source. 

Im thinking if you have impact levels and you don't specify a target you'll get the FedRAMP result and the impact level result.

meganwolf0:
> I noticed the 4 findings but I think it's a good thing.
> 
> You'll get one per framework and one per catalog/source.
> 
> Im thinking if you have impact levels and you don't specify a target you'll get the FedRAMP result and the impact level result.

Would there be a situation where source and framework would be different or like independent? They just seemed to kind of indicate the same thing but different values, basically. But there might be more nuance that the example doesn't really cover

Edit: Talked to Brandt, I get it :) disregard this one

brandtkeller:
> A couple other questions/comments as I'm reviewing:
> 
> * Looks like observations aren't all appended to `related-observations` when two different components are calling the same control (perhaps we need a test case for this scenario)

This is a real bug - should be fixed by checking for existing related observations instead of overwriting when generating findings. 

> * This one is more personal preference - but the "Findings" section in the validate output is taking up a lot of space in the terminal, especially just with line breaks, would it be possible to condense those to a single line -> Control ID: %s, Finding UUID: %s, Status: %s? Maybe there's a reason they should be that expansive though so happy to hear the counterargument there.

Agreed - I re-used the table functionality from message and consolidated the output to `Control ID / Status`. Open to feedback on if we really need the finding UUID (which are really just unique control ID's to my knowledge)

> * How does `framework` get added? It appears to be manual right now, is that something that ideally will happen via `lula generate`?

Manually or as part of generate in another PR: #574

meganwolf0 (approved):

Issue #538, opened 2024-07-15
OSCAL Lint Error on Composed Test OSCAL File


CloudBeard:
### Environment
Device and OS:  Ubuntu 20.04
App version: v0.4.2

### Steps to reproduce
1. `lula tools compose -f src/test/unit/common/composition/component-definition-import-compdefs.yaml `
2. `lula tools lint -f src/test/unit/common/composition/component-definition-import-compdefs-composed.yaml`

### Expected result
` ✔  Linting src/test/unit/common/composition/component-definition-import-compdefs-composed.yaml`
### Actual Result
```shell
  •  Validation result for                                                                                                                                                                                        
       src/test/unit/common/composition/component-definition-import-compdefs-composed.yaml: { "valid":
       false, "timeStamp": "2024-07-15T14:20:07.806777522Z", "errors": [ { "keywordLocation":
       "/oneOf/0/required", "absoluteKeywordLocation":
       "http://csrc.nist.gov/ns/oscal/1.0/1.1.2/oscal-complete-schema.json#/oneOf/0/required",
       "instanceLocation": "", "error": "missing properties: 'catalog', missing properties: 'profile'" }, {
       "keywordLocation":
       "/oneOf/2/properties/component-definition/$ref/properties/back-matter/$ref/properties/resources/items/properties/uuid/$ref/pattern",
       "absoluteKeywordLocation":
       "http://csrc.nist.gov/ns/oscal/1.0/1.1.2/oscal-complete-schema.json#/definitions/UUIDDatatype/pattern",
       "instanceLocation": "/component-definition/back-matter/resources/1/uuid", "error": "does not match
       pattern
       '^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[45][0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$'",
       "failedValue": "123e4567-e89b-12d3-a456-426614174000" }, { "keywordLocation": "/oneOf/3/required",
       "absoluteKeywordLocation":
       "http://csrc.nist.gov/ns/oscal/1.0/1.1.2/oscal-complete-schema.json#/oneOf/3/required",
       "instanceLocation": "", "error": "missing properties: 'system-security-plan', missing properties:
       'assessment-plan', missing properties: 'assessment-results', missing properties:
       'plan-of-action-and-milestones'" } ], "metadata": { "documentPath":
[...]

PR #539, opened 2024-07-15
fix(test): updated uuid in kyverno validation

CloudBeard:
## Description

The Kyverno validation UUID in the dev-validate e2e folder was an incorrect UUID. It needed to start with a 4 in the 3rd set of characters and had an extra character at the end.

The Composition test uses the remote location of  `https://raw.githubusercontent.com/defenseunicorns/lula/main/src/test/e2e/scenarios/dev-validate/validation.kyverno.yaml` to get the validation during compose so the composed artifact is invalid because of the UUID of the validation.

The bad UUID caused the error when the file is linted.

By changing the UUID it fixes the issues in the composed artifact which fixes the test that is ran by the compose_test.go which composed the OSCAL files in the src/test/unit/common/composition test directory.

## Related Issue

Fixes #
https://github.com/defenseunicorns/lula/issues/538 by updating the incorrect UUID in the file used for testing.

## Type of change

## Checklist before merging


CloudBeard:
E2E will fail until this fix is pushed to main.

https://github.com/defenseunicorns/lula/blob/575500e0ce40f786d2b6cdf143757e10e6e2c333/src/test/unit/common/composition/component-definition-local-and-remote.yaml#L43

The compose is testing local and remote OSCAL and this component-definition is looking at main and this change isn't there yet. Should be updated once and good to go.

Maybe pin a version once fixed and track with renovate? I think main is fine, shouldn't change much.

brandtkeller:
> E2E will fail until this fix is pushed to main.
> 
> https://github.com/defenseunicorns/lula/blob/575500e0ce40f786d2b6cdf143757e10e6e2c333/src/test/unit/common/composition/component-definition-local-and-remote.yaml#L43
> 
> The compose is testing local and remote OSCAL and this component-definition is looking at main and this change isn't there yet. Should be updated once and good to go.
> 
> Maybe pin a version once fixed and track with renovate? I think main is fine, shouldn't change much.

There is a gap here for testing against `main` resources and doing feature development. Someone may unknowingly introduce a change to a file that is not picked-up until merge to main has happened. 

Some of this can be resolved with ensuring we have insights to valid testing files actually being valid (linted etc) and remaining that way. We should file an issue that looks at some method for classifying a file as a valid artifact (possibly filename identifier) and performs some linting to ensure they stay valid.

src/test/unit/common/composition/temp.yaml
```diff
@@ -0,0 +1,50 @@
+assessment-results:
```
brandtkeller: Was this meant to make it into the changes?

brandtkeller:
Appreciate jumping to get this fixed. Only other comment I would make is additional context to assist reviews; you've got:
- What is the problem
- Where does it exist
- Why does it exist

The only missing piece of value is 'How to validate the changes'.

src/test/unit/common/composition/temp.yaml
CloudBeard: That needs removed. From a separate test that shouldn't be on this branch.

mike-winberry (approved):

CloudBeard:
Verification of fix.

First screenshot is the fail in the E2E testing GitHub workflow. The test that is failing is the TestComponentDefinition in the `src/cmd/tools/compose_test.go`
 
image

Fix screenshots below show the `src/test/unit/common/composition/component-definition-local-and-remote.yaml` changed the url to this pr branch as the remote source and re-running the test specifically. (had some issues running make test but I believe thats something on machine vs the actual environment, more failing there than in CI)

image

brandtkeller (approved):

PR #540, opened 2024-07-15
feat(evaluate): add observation details

### LINKED ISSUE #533: Additional information on Evaluate
When evaluate runs, and fails, only the control IDs are output. More information would be helpful for debugging what actually failed, since the control satisfaction is often a function of many observations.

- **Given** assessment-results
- **When** `lula evaluate` fails
- **Then** list observation changes

### Additional context
Thinking a decent solution would be:
- Identify old -> new observation pairs by the observation.description text matching. Print out those with mis-matched `relevant-evidence.description` values (i.e., the result: not-satisfied/satisfied)
- For any new observations that are not-satisfied without a previous old observation, print those.

This is just my idea for evaluate, but perhaps this rolls into a larger function to interrogate the assessment results more granularly and specific to individual controls.

meganwolf0:
## Description

Adds some additional detail to the `evaluate` command when findings between New and Threshold assessments are different. Ideally this will give the person enough information to quickly know what changes changed the satisfaction of the findings.

Added a `--summary` flag to evaluate to print all observations.

Current output for changed assessments is in a table format for all observations so as to not duplicate single observations that fail on multiple controls.

## Related Issue

Fixes #533 

## Type of change

## Checklist before merging


meganwolf0:
Sample view for the user - shows both sets of results leading to failing eval + "summary" to get all observations. I think these can be rearranged slightly and can have more fidelity over the info returned to try and not overwhelm the user, but wanted to throw out for spears

Screenshot 2024-07-17 at 3 23 32 PM

src/cmd/evaluate/evaluate.go
```diff
@@ -97,19 +120,50 @@ func EvaluateAssessments(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentRe
...
+				"removed-not-satified": resultComparison["removed-not-satified"],
+			}
+			findingsWithoutObservations = result.Collapse(failedFindings).PrintObservationComparisonTable(true, false, true)
+			// handle controls that failed but didn't have observations
+			if len(findingsWithoutObservations) > 0 {
+				message.Warnf("%d Failed Finding(s) Without Observations", len(findingsWithoutObservations))
+				message.Info(strings.Join(findingsWithoutObservations, ", "))
 			}
+
+			// Print by individual table
```
meganwolf0: Left this block in as an alternative printing style to above, could always put this under a flag or just remove entirely if the more consolidated table output is preferred.

src/cmd/evaluate/evaluate.go
```diff
@@ -69,22 +73,41 @@ func EvaluateAssessments(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentRe
...
 		message.Debugf("threshold UUID: %s / latest UUID: %s", resultMap["threshold"].UUID, resultMap["latest"].UUID)
 
-		status, findings, err := oscal.EvaluateResults(resultMap["threshold"], resultMap["latest"])
+		status, resultComparison, err := oscal.EvaluateResults(resultMap["threshold"], resultMap["latest"])
 		if err != nil {
 			message.Fatal(err, err.Error())
 		}
 
+		// Print summary
+		if summary {
```
meganwolf0, line 89: This "summary" is really just a table of the observations. We could probably easily add more data here if we wanted to... planning to update the `evaluate` command in the docs with --summary info if this seems ok.

src/pkg/common/oscal/assessment-results.go
```diff
@@ -177,58 +170,83 @@ func IdentifyResults(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentResult
...
+			status:      true,
+		},
+		{
+			name:        "unchanged-satisfied",
+			stateChange: result.UNCHANGED,
+			satisfied:   true,
+			status:      true,
+		},
+		{
+			name:        "removed-not-satified",
```
brandtkeller:
```suggestion
			name:        "removed-not-satisfied",
```

src/pkg/common/oscal/assessment-results.go
```diff
@@ -177,58 +170,83 @@ func IdentifyResults(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentResult
...
+			stateChange: result.REMOVED,
+			satisfied:   false,
+			status:      false,
+		},
+		{
+			name:        "removed-satified",
```
brandtkeller:
```suggestion
			name:        "removed-satisfied",
```

src/cmd/evaluate/evaluate.go
```diff
@@ -97,19 +120,50 @@ func EvaluateAssessments(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentRe
...
-			message.Warn("Evaluation Failed against the following findings:")
-			for _, finding := range findings["no-longer-satisfied"] {
-				message.Warnf("%s", finding.Target.TargetId)
+			// Print no-longer-satisfied
+			message.Warn("Evaluation Failed against the following:")
+
+			// Alternative printing in a single table
+			failedFindings := map[string]result.ResultComparisonMap{
+				"no-longer-satisfied":  resultComparison["no-longer-satisfied"],
+				"removed-satified":     resultComparison["removed-satified"],
```
brandtkeller:
```suggestion
				"removed-satisfied":     resultComparison["removed-satisfied"],
```

src/cmd/evaluate/evaluate.go
```diff
@@ -97,19 +120,50 @@ func EvaluateAssessments(assessmentMap map[string]*oscalTypes_1_1_2.AssessmentRe
...
+				"removed-not-satified": resultComparison["removed-not-satified"],
```
brandtkeller:
```suggestion
				"removed-not-satisfied": resultComparison["removed-not-satisfied"],
```

brandtkeller:
The information at play here makes the workflow very observation-centric. I don't see anything wrong with that and this is a huge improvement for exposing information. 

I would be curious if you have thoughts for the output and the Observation to Finding relationship?

IE if a control that was previously passing now has multiple failing observations - this results in multiple line-items in the output - because multiple observations are failing. Ultimately these multiple failing observations roll-up to a single failing finding (for which has an actual field for capturing state). 

Example (ignore the new/threshold remarks columns):
```bash
dev@dev:~/work/lula$ ./bin/lula evaluate -f assessment-results.yaml

 NOTE  Saving log file to /tmp/lula-2024-07-18-23-11-19-4230766809.log
  ⠋  Evaluating Assessment Results 5a0a9538-e734-48a5-a327-02e6aa6891b0 against b57d4be5-75a7-4b3b-be66-e3cfcb1cac15 (0s)
 WARNING  Evaluation Failed against the following:                                                                                                                                                                                                                                                                                                                             

Control ID(s) | Observation                                                | Satisfied | Change                     | New Remarks                                                         | Threshold Remarks                                                                                                                                                                  
sc-23         | 67456ae8-4505-4c93-b341-d977d90cb125 - istio-health-check  | false     | SATISFIED TO NOT SATISFIED | istiohealth.deployment_message: All deployment conditions are true. | istiohealth.deployment_message: All deployment conditions are true.
              |                                                            |           |                            | istiohealth.hpa_message: HPA has sufficient replicas.               | istiohealth.hpa_message: HPA has sufficient replicas.
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
sc-23         | ca49ac97-487a-446a-a0b7-92b20e2c83cb - enforce-mtls-strict | false     | SATISFIED TO NOT SATISFIED | validate.msg: All PeerAuthentications have mtls mode set to STRICT. | validate.msg: All PeerAuthentications have mtls mode set to STRICT.
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

     ERROR:  failed to meet established threshold 
```

Current approach works well for the summary containing the state of observations and those mapped to many Findings (control-ids really) - but I wonder if we need to augment or add something in the future that is more Finding-centric?

meganwolf0:
> Current approach works well for the summary containing the state of observations and those mapped to many Findings (control-ids really) - but I wonder if we need to augment or add something in the future that is more Finding-centric?

Yeah I think this would be totally do-able. I do a kind of refactor of the paired results to get the observations by controls, so reporting out more info about the original object would probably be a simple thing to add. Probably would need to talk more about what would be best to try and tease out of that info for the user? [This alternate table](https://github.com/defenseunicorns/lula/pull/540#discussion_r1683401830) tried to just report each control along with the observations.. so you could kind of slice it that way too. I just didn't love that for the scenario where you had like a single observation that's attached to a bunch of different findings failing.

src/cmd/evaluate/evaluate.go
brandtkeller: Looks good to me. Provides an entry-point for future enhancements via opt-in flag.

meganwolf0: added the details to the `evaluate.md` doc

src/cmd/evaluate/evaluate.go
brandtkeller: Current Output provides steps forward for people to evaluate. No demand signals currently for all or both so maybe we remove and evaluate going forward?

docs/cli-commands/assessments/evaluate.md
```diff
@@ -2,6 +2,28 @@
...
+```
+
+To print a summary of the observation results:
+```bash
+lula evaluate -f assessment-results.yaml --summary
+```
+
+## Options
+
+- `-f, -file`: The path to the file(s) to be evaluated.
```
brandtkeller:
```suggestion
- `-f, --file`: The path to the file(s) to be evaluated.
```

brandtkeller (commented):
Quick cleanup items and I believe this is ready. One of the first instances of improving UX and feedback will probably be the main driver for updates outside of general assumptions.

src/cmd/evaluate/evaluate.go
meganwolf0: removed

CloudBeard (approved):

brandtkeller (approved):

PR #553, opened 2024-07-23
fix(oscal): deterministic OSCAL model write

### LINKED ISSUE #542: Deterministic OSCAL Writes
When lula performs a write (whether in isolation or with a merge) the models use of arrays lead to random ordering of objects. 

As lula encourages the version control of OSCAL and the review of changes - it will become more pertinent that the changes reflect the modified lines and not mass-updates due to changing order of existing/not-modified items. 

- **Given** an operation that performs oscal writes
- **When** an oscal file is generated
- **Then** the order for performing the same action is the same

No Alternatives

### Additional context
Each model needs consideration for array use and performing a sort on some key. 

`lula generate component` may be a good starting point given the easy identification of many nested objects.

brandtkeller:
## Description

This fix implements sorting based on the fields we currently operate within on `component-definition` and `assessment-results` to  make determinism both an initial write activity as well as a corrective action.

Goal for OSCAL operations is to not assume anyone has always started with Lula - rather allowing for people to bring their OSCAL and enable Lula to provide value without strong opinionation. This is why the execution is centered on the write as opposed to creation of objects. This makes it applicable any time a file is written to - including compose, validate, evaluate, generate. 

I see this change as important for those storing resources in version control. The `component-definition.md` document did not exist - So i've started an initial draft for iteration. we should make this a requirement each time a new model is introduced.

Please check my process - if there are more optimal ways to accomplish this I am all ears/eyes. 

Edit: also added a cspell.json file to the repository to start tracking spelling because I apparently cannot spell `satisfied` 

## Related Issue

Fixes #542

## Type of change

## Checklist before merging


docs/oscal/component-definition.md
```diff
@@ -0,0 +1,16 @@
+# Component Definition
+
+a [Component Definition]() is an OSCAL model for capturing control information that pertains to a specific component/capability of a potential system. It can largely be considered the modular and re-usable model for use across many systems. In Lula, the `validate` command will process a `component-definition`, iterate through all `implemented-requirements` to discover Lula validations, and execute those validations to produce `observations`.
```
meganwolf0:
```suggestion
A [Component Definition](https://pages.nist.gov/OSCAL/resources/concepts/layer/implementation/component-definition/) is an OSCAL model for capturing control information that pertains to a specific component/capability of a potential system. It can largely be considered the modular and re-usable model for use across many systems. In Lula, the `validate` command will process a `component-definition`, iterate through all `implemented-requirements` to discover Lula validations, and execute those validations to produce `observations`. 
```

meganwolf0:
`istio-composed.yaml` is probably there accidentally, just calling it out

src/pkg/common/oscal/component.go
```diff
@@ -432,6 +433,78 @@ func ComponentDefinitionToRequirementMap(componentDefinition *oscalTypes_1_1_2.C
 	return requirementMap
 }
 
+func MakeComponentDeterminstic(component *oscalTypes_1_1_2.ComponentDefinition) {
+	// sort components by title
+
+	if component.Components != nil {
+		components := *component.Components
+		sort.Slice(components, func(i, j int) bool {
+			return components[i].Title < components[j].Title
```
meganwolf0, line 442: Appears that this sort method is case sensitive (uppercase is placed before lowercase). I think this is probably fine, just calling it out so that we are aware of the case sensitive ordering if we want to try and establish some convention there.

src/pkg/common/oscal/assessment-results.go
```diff
@@ -231,6 +232,56 @@ func EvaluateResults(thresholdResult *oscalTypes_1_1_2.Result, newResult *oscalT
...
+	// Sort Results
+	slices.SortFunc(assessment.Results, func(a, b oscalTypes_1_1_2.Result) int { return b.Start.Compare(a.Start) })
+
+	// Sort Targets
+	for _, result := range assessment.Results {
+		// sort findings by target id
+		if result.Findings != nil {
+			findings := *result.Findings
+			sort.Slice(findings, func(i, j int) bool {
+				return findings[i].Target.TargetId < findings[j].Target.TargetId
```
meganwolf0, line 263: I think this meets intent, but this ordering is kinda wonky since it'll put something like AC-13 before AC-3, or SC-7.21 before SC-7.4. Maybe a more specific sort would be useful at some point for the control IDs to properly sort on the numerical portion.

src/pkg/common/oscal/assessment-results.go
```diff
@@ -231,6 +232,56 @@ func EvaluateResults(thresholdResult *oscalTypes_1_1_2.Result, newResult *oscalT
...
+			})
+			result.Findings = &findings
+		}
+		// sort observations by collected time
+		if result.Observations != nil {
+			observations := *result.Observations
+			slices.SortFunc(observations, func(a, b oscalTypes_1_1_2.Observation) int { return a.Collected.Compare(b.Collected) })
```
meganwolf0, line 270: This doesn't appear to be deterministic, we could possibly sort on `description`? I think that at least is a function of the validation uuid + name, so ideally won't really be changing across runs

src/cmd/tools/compose.go
```diff
@@ -81,7 +78,7 @@ func Compose(inputFile, outputFile string) error {
 	}
 
 	// Write the composed OSCAL model to a file
-	err = WriteComposedOscalModel(model, outputFile, inputFile)
+	err = oscal.WriteOscalModel(outputFile, model)
```
meganwolf0, line 81: I have sort of generic complaint about the way `compose` handles missing output file names versus `validate` - here we're getting the full path of the input file and putting the composed file there, but in validate we're putting the file just in the working directory (seems that oscal.WriteOscalModel has a function to determine empty filePath). Small nit, not sure it needs to go in this PR, but I saw it and wanted to highlight the discrepancy.

src/pkg/common/oscal/assessment-results.go
meganwolf0: Another really small issue here is that the `remarks` in the observations may also re-order... that's in the validation-store RunValidations function where the val.Result.Observations are added as remarks... that may want for sorting at some point too.

src/cmd/tools/compose.go
brandtkeller: agreed - behavior is different and that sounds like a good issue to standardize going forward. IE "where do default file writes occur? "

src/pkg/common/oscal/assessment-results.go
brandtkeller: I think that is a fair argument - the only outlier I see is that `target-id` can be anything. Yes it's `AC-13`, `AC-3` in 800-53 but may be `Banana` and `Orange` in some health regulatory standard. (Exaggerated for dramatic effect).

src/pkg/common/oscal/assessment-results.go
brandtkeller: Agree on the remarks - that sort might be difficult as it's all one field.... Need to think on that more. 

What about the use of `Collected` causes inconsistent ordering? is it a by-product of validations ran in a random order?

src/pkg/common/oscal/component.go
brandtkeller: Agreed! I will make a note in the docs.

src/cmd/tools/compose.go
brandtkeller: [Issue](https://github.com/defenseunicorns/lula/issues/560) created

CloudBeard (approved):
Changes look good to me. Tested locally with component-definition and assessment-results and get expected results.

Im not aware of a better way to sort. 

Since its machine readable I think the main sorting/ordering for the look could be done in reporting/data gathering.

src/pkg/common/oscal/assessment-results.go
meganwolf0: Yeah I believe it's due to those being run randomly, so the timestamps don't always appear in the same order... so that could probably be sorted?

brandtkeller: Discussion for posterity:
> I was thinking about the separation between how the validations are executed (and determinism) vs the activity of writing an assessment-result and I don't think the impact is too high either way.
IE a result is a historical artifact - the contents of a result shouldn't change after it's written but it would be nice to ensure the contents of whatever was written doesn't change over time (or through other read/write activities).
Doesn't invalidate that the execution isn't necessarily deterministic

src/pkg/common/oscal/assessment-results.go
meganwolf0: makes sense to me!

meganwolf0 (approved):

PR #568, opened 2024-07-28
feat: yaml map injection

### LINKED ISSUE #567: Generic yaml merge functionality
To facilitate generation of more complex OSCAL models, e.g., SSP, we need a functionality to inject yaml maps or sequence data into a target.

This issue is just coverage for that generic map merge functionality that may be re-used in various OSCAL model combination use cases.

meganwolf0:
## Description

Generic library function to inject a "subset" map into a "target" map. Merging supports appending lists instead of replacing them, as it was assumed that appending to an existing list would be more appropriate but replacement could be supported if we wanted to as well.

Updated to add OSCAL map merge as well to paint a broader picture of the intended use case

## Related Issue

Fixes #567 

## Type of change

## Checklist before merging


brandtkeller:
Tests helped provide some basic insights into intent. 

Thinking about where we might be able to put this into use - I imagine, as discussed previously, that this would be for common config merging into OSCAL models. From a sequence perspective -> we have an OSCAL model + some config -> we read the model and config into maps -> perform the injection -> convert to bytes and then marshall to oscal model? 

Given the map parameters I imagine we can support json in an intended workflow?

meganwolf0:
> Thinking about where we might be able to put this into use - I imagine, as discussed previously, that this would be for common config merging into OSCAL models. From a sequence perspective -> we have an OSCAL model + some config -> we read the model and config into maps -> perform the injection -> convert to bytes and then marshall to oscal model?

Yep, that's exactly what I was thinking!

> Given the map parameters I imagine we can support json in an intended workflow?

Also, yep JSON should work

brandtkeller (dismissed):
Created #624 to track future implementation of this functionality. Does not necessarily require the templating decisions given this is separate scope and the viper configuration is already present.

brandtkeller (changes requested):
determine package naming and evaluate placement in `src/internal` for lula internal use only.

meganwolf0:
> determine package naming and evaluate placement in `src/internal` for lula internal use only.

changed package name to `inject`, moved to `internal`. Also happy to call `merge` if we think `inject` doesn't make a ton of sense... there's just a lot of "merges" out there

brandtkeller (approved):

PR #585, opened 2024-08-07
fix(validate): get non-namespace scoped resources

### LINKED ISSUE #409: Can't get non-namespace scoped resource by name
### Environment
Device and OS:  
App version:    
Kubernetes distro:  k3s
Kubernetes version:  1.27
provider:
- [ ] kyverno
- [ ] opa

### Steps to reproduce
1. Request a named resource that is non-namespace scoped

### Expected result
Should return resource

### Actual Result
Does not return resource

### Additional Context
The `GetResourcesDynamically` function requires that a named resource have a namespace, but this is probably not a good assumption when trying to retrieve non-namespaced cluster resources. Unsure if a new flag to indicate non-namespace scope is appropriate to remove the dual usage of the namespace field (i.e., if it's empty get all resources as well as it denotes non-namespaced) or just allow a named resource to be retrieved from any namespace.

meganwolf0:
## Description

Updates some of the logic to identify which resources are cluster-scoped, and allow those that are to be reduced by name.

## Related Issue

Fixes #409

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging


brandtkeller:
I am appreciative of the thoroughness for checking if the resource is namespaced. 

While I am likely not considering some edge cases you have ran into - I want to play devils advocate for a second. a non-namespaced resource must have a unique name for all same resources. Given a resource name being defined in the kubernetes domain resource item - this is a non-problem (to my understanding) - as we can assume a non-namespaced & named resource would be unique. 

but the problem here lies in namespaced items not being unique across the cluster - could we instead change `reduceByName` to collect all instances and error when > 1? preventing additional network activity for ever single resource I believe?

meganwolf0:
> could we instead change `reduceByName` to collect all instances and error when > 1?

I think yeah, this would work. I think perhaps this change in logic would just result in less error fidelity - for instance:
If a user requests a named pod and no-namespace specified - In one case if there is only one pod matching the name in the cluster, they will get a result. If for some reason the cluster changes and now there's two pods, same name, different namespaces, then the same request will result in an error. If we had know that "pod" should require a namespace the first time around, then that resource-rule would have been blocked properly. 

I don't love the request to the server to to find if something is namespace scoped, so I think it's probably just a tradeoff that could go either way.

The only other things I could think of to solve this would be breaking, but some other options could be:
- request `cluster-scoped` tag (probably default to false so most validations wouldn't need to add it)
- return the full list of all matching named things, regardless of namespace - also breaking since we currently only assume one item is returned

meganwolf0:
I think actually the third alternative here is to just use the Get method on the dynamic client instead of list, as that seems to handle all of this natively... will explore and update the PR

brandtkeller:
Agreed on the tradeoffs discussion - worth framing through the lens of "which would we want to support in production?" 

Which behaviors are going to require the least amount of adaptation in the event of failure. A validation requiring update could be annoying yet would provide a more stable validation over time. 

> return the full list of all matching named things, regardless of namespace - also breaking since we currently only assume one item is returned

Another flavor of this I am still noodling on for implications would be the idea of returning a list if many or a single if one (given the generic interface).

meganwolf0:
Ok I updated as follows
- Removed the `reduceByName` and `isClusterScoped` functions in favor of dynamic's `Get` - as mentioned this seems like the more native solution
- Updated the test cases, both adding ones that should fail and one for a cluster-scoped resource request
- Updated the schema - I removed the rule that namespaces should be specified if name is for resource-rule - if a namespaced resource is requested without a namespace, the `Get` will fail anyway, so not sure this needs to be in the schema since for cluster-scoped ones it's an invalid assumption. I also had to make the type one of "array" or "null" to handle where namespaces is just not set at all. I think that was a gap before because it always failed when it wasn't defined, regardless of if it was actually needed. I don't believe this breaks anything but could be missing something.

src/pkg/domains/kubernetes/resource.go
```diff
@@ -66,38 +66,42 @@ func GetResourcesDynamically(ctx context.Context,
...
-		if err != nil {
-			return nil, err
-		}
-
-		// Reduce if named resource
+	if resource.Name != "" && len(namespaces) > 1 {
+		return nil, fmt.Errorf("named resource requested cannot be returned from multiple namespaces")
+	} else if resource.Name != "" && len(namespaces) == 1 {
+		// Extracting named resources can only occur here
 		if resource.Name != "" {
```
brandtkeller: Given the outer conditional is this `if resource.Name != ""` still required?

meganwolf0: I think both `if resource.Name != ""` are needed, perhaps we can remove the second `len(namespaces)==1` since that should be the default if the len(namespaces) is not >1

brandtkeller: for posterity - we're both talking about the same conditional [here](https://github.com/defenseunicorns/lula/pull/585/files#diff-3e52c5faf4292c7b17bafab907db20fe830c09884dc91775a1c654b318895ef1R79) correct?

meganwolf0: Oh sorry yeah missed that one - that is redundant

brandtkeller (changes requested):
Changes look good - two considerations for feedback:

- Docs are mostly still accurate - but only mention named resource use when targeting a specific namespace. A mention of cluster scoped resources might be good. 
- Quick augmentation of the tests for the added failure validations would be nice to check the observations for specific errors versus simply `not-satisfied` findings.

meganwolf0:
> * Quick augmentation of the tests for the added failure validations would be nice to check the observations for specific errors versus simply `not-satisfied` findings.

What are you thinking for this one? My initial thought is that you're wanting me to validate all the remarks are equal to some string for each of the failure cases? If thats the case this may require more than a little rework, just checking before I go forth

brandtkeller:
> > * Quick augmentation of the tests for the added failure validations would be nice to check the observations for specific errors versus simply `not-satisfied` findings.
> 
> What are you thinking for this one? My initial thought is that you're wanting me to validate all the remarks are equal to some string for each of the failure cases? If thats the case this may require more than a little rework, just checking before I go forth

That was my thought but open to that being a separate issue if necessary. I see a variety of different scenarios in the failure cases (which is good) that have resulted in findings being listed as `not-satisfied` but the observations range from errors at various different points and some Validations being evaluated and others not-evaluated.

It would be nice to trace a known error handling scenario is actually hitting the intended handling target.

meganwolf0:
> It would be nice to trace a known error handling scenario is actually hitting the intended handling target.

Yeah I agree, I don't love that they're all not satisfied without knowing the reason... I also feel like checking the remarks text is also potentially a little too blunt an instrument. I think maybe having like an error code returned and checking that might be better? But yeah I would probably say that should be a separate issue and would require a little bit of time to architect properly

brandtkeller (approved):
Great call looking further upstream in the sequence to find a best place to resolve the issue.

PR #598, opened 2024-08-14
refactor(tests): 586 refactor unhappy path e2e tests

### LINKED ISSUE #586: refactor unhappy path e2e tests

Currently the unhappy paths for validations are in a single [oscal file](https://github.com/defenseunicorns/lula/blob/main/src/test/e2e/scenarios/pod-label/oscal-component-all-bad.yaml) that results in all findings `not-satisfied`. This isn't a very robust check since `not-satisifed` is too generic to tell us why it's not-satisfied, which is ideally what we should be validating. 

test file: https://github.com/defenseunicorns/lula/blob/main/src/test/e2e/pod_validation_test.go

### Additional context

Perhaps creating an error code or similar to tell lula why an observation is not satisfied. Additionally, some of these may be best pulled out into unit tests where applicable (e.g., for linting failures)

meganwolf0:
## Description

Overall intent is to make sure all "unhappy path" testing is done for the validations and `lula dev validate`. Much of this is caught by the schema, some of it is not. Added tests for individual domains/providers assuming they can be consumed in absence of schema validation methods. The following is a list of updates/additions, working from the most narrow-scope upward:

- Updated functionality in each domain and provider to create/validate the spec structure and return errors if invalid + unit tests to exhaustively document valid/invalid cases for each
- Updated GetDomain and GetProvider testing to for error handling/evaluation
- Added common/types testing for ToLulaValidation to check for all error types 
  -> Required updated schema -> opa-spec.output.observations is allowed to be nil
- Added unhappy path tests for composition unit and e2e tests, this was pulled out of the `src/test/e2e/scenarios/pod-label/oscal-component-all-bad.yaml`
- Updated `pod_validation_test.go` to map the expected error type to the specific observation failure

## Related Issue

Fixes #586 

## Type of change

- [x] Other (addresses tech debt in testing, slight schema modification)

## Checklist before merging


src/pkg/common/composition/composition_test.go
```diff
@@ -30,6 +32,16 @@ func TestComposeFromPath(t *testing.T) {
 		}
 	})
 
+	t.Run("No imports, local validations, bad href", func(t *testing.T) {
```
brandtkeller, line 35: These two new tests are just happy path tests with known bad conditions?

src/pkg/common/types_test.go
```diff
@@ -0,0 +1,189 @@
...
+provider:
+  type: "opa"
+  opa-spec:
+    rego: "package validate\n\ndefault validate = false"
+`),
+			expectErr:       true,
+			expectedErrType: common.ErrInvalidSchema,
+		},
+		{
+			name: "Invalid provider scheam, bad type",
```
brandtkeller:
```suggestion
			name: "Invalid provider schema, bad type",
```

src/pkg/common/types_test.go
```diff
@@ -0,0 +1,189 @@
...
+  kubernetes-spec:
+    resources: []
+provider:
+  type: "unknown"
+`),
+			expectErr:       true,
+			expectedErrType: common.ErrInvalidSchema,
+		},
+		{
+			name: "Invalid provider scheam, missing spec",
```
brandtkeller:
```suggestion
			name: "Invalid provider schema, missing spec",
```

src/pkg/domains/kubernetes/types.go
```diff
@@ -15,6 +16,77 @@ type KubernetesDomain struct {
...
 }
 
+func CreateKubernetesDomain(ctx context.Context, spec *KubernetesSpec) (types.Domain, error) {
+	// Check validity of spec
+	if spec == nil {
+		return nil, fmt.Errorf("spec is nil")
+	}
+
+	if spec.Resources == nil && spec.CreateResources == nil && spec.Wait == nil {
+		return nil, fmt.Errorf("resources, create-resources, or wait must be specified")
```
brandtkeller:
```suggestion
		return nil, fmt.Errorf("one of resources, create-resources, or wait must be specified")
```

brandtkeller (changes requested):
Minor questions and typos otherwise I very much appreciate this baseine.

src/pkg/common/composition/composition_test.go
meganwolf0: Yeah exactly - I had some bad href values in the `src/test/e2e/scenarios/pod-label/oscal-component-all-bad.yaml` that I felt should be over here

brandtkeller: And to clarify - the purpose of these tests is to ensure these scenarios do not return any errors?  Should they be returning errors given the bad inputs?

meganwolf0: I think it's supposed to be composing cleanly - i.e., no errors... I think that's sort of how we've been operating such that those errors get raised up on validate (in the remarks), instead of err-ing during compose... so I think the result of this test tracks with how we've been using this function, idk if we want to revisit that though

brandtkeller: Fair!

brandtkeller (approved):

PR #599, opened 2024-08-14
feat: initial lula report

### LINKED ISSUE #223: Reporting of number of controls met by component definitions

CloudBeard:
# Description

Adds report functionality starting with component definition.

Will report how many controls are mapped per source and per framework for each component. There is the possibility that a component-definition file has multiple sources and multiple frameworks. The current report will list how many controls are in each and will give a % mapped based on the source provided for each.

Running `lula generate component --catalog-source https://raw.githubusercontent.com/GSA/fedramp-automation/master/dist/content/rev5/baselines/json/FedRAMP_rev5_MODERATE-baseline-resolved-profile_catalog.json --component 'Keycloak' --requirements ac-12,ac-2,ac-2.1,ac-2.3,ac-2.4,ac-2.5,ac-2.7,ac-3,ac-5,ac-6,ac-6.1,ac-6.2,ac-6.5,ac-6.9,ac-6.10,ac-7,ac-8,au-2,au-3,au-3.1,au-8,au-9.4,au-12,cm-5.1,cm-5.5,ia-2,ia-4.4,ia-5.1,ia-5.2,ia-6,ia-8.1,ia-11,sc-13 --remarks assessment-objective --framework il4`

Will show that ac-2 is fixed but also show that other controls aren't broken as a side effect of the fix.

## Related Issue

Relates to #223 and #247 

## Type of change

## Checklist before merging


brandtkeller:
Quick feedback loop - I can trace the code but I would always recommend starting with docs on a new feature. 

Capture the purpose - let people challenge that intent. 

I don't believe this needs an ADR - so docs would be sufficient. 

At the same time I think we should keep challenging if we know enough to pursue it further. Does this provide enough value to be of use or does some infrastructure require more maturation first. Transparently I do not know if we have enough context to make the result meaningful.

CloudBeard:
Still need to add tests but want to figure out where/what packages to create to tidy up the main report code.

src/cmd/report/report.go
```diff
@@ -0,0 +1,281 @@
...
+`
+
+var reportCmd = &cobra.Command{
+	Use:     "report",
+	Hidden:  false,
+	Aliases: []string{"r"},
+	Short:   "Build a compliance report",
+	Example: reportHelp,
+	Run: func(_ *cobra.Command, args []string) {
+		if opts.InputFile == "" {
```
brandtkeller: If we mark this flag as [required](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#required-flags) then we can remove this check

src/cmd/report/report.go
```diff
@@ -0,0 +1,281 @@
...
+
+type ComponentDefinitionReportData struct {
+	Title                string            `json:"title" yaml:"title"`
+	ControlIDBySource    map[string]int    `json:"control ID mapped" yaml:"control ID mapped"`
+	ControlIDByFramework map[string]int    `json:"controlIDFramework" yaml:"controlIDFramework"`
+}
+
+var opts = &flags{}
+
+var reportHelp = `
```
brandtkeller, line 11: Capture the full intent of what this command will perform here (different options etc). Later doc generation (prior to review) will recreate the docs above.

src/cmd/report/report.go
```diff
@@ -0,0 +1,281 @@
...
+
+		spinner = message.NewProgressSpinner("Reading OSCAL model from file")
+		oscalModels, err := readOSCALModel(getOSCALModelsFile)
+		if err != nil {
+			spinner.Fatalf(err, "Failed to get OSCAL Model data")
+		}
+		spinner.Success()
+
+		spinner = message.NewProgressSpinner("Determining OSCAL model type")
+		modelType, err := determineOSCALModel(oscalModels)
```
brandtkeller: This may be a good spot to establish a separation of duties -> We've performed the essential operation of retrieving/parsing the data -> now let's create a function that handles the next steps.

src/cmd/report/report.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			spinner.Fatalf(err, "Unable to determine OSCAL model type")
+		}
+		spinner.Success()
+
+		spinner = message.NewProgressSpinner("Generating report")
```
brandtkeller: I don't believe the intent of this operation is going to be an OR but rather an AND in the long term. Can we possibly re-orient functionality to process the context that is present? (focusing on component-definitions here)

brandtkeller (changes requested):
Trending in the right direction now that we're getting to core of what we need (or can provide currently). 

I think we're still missing a bit of big picture vision -> Which I/the team can help define. 

See my comment about the context that drives the reporting (as we see it today). Ideally we as a team need to provide a clearer deliverable on our issues (WIP) - but I appreciate you pursuing this nonetheless. 

My current take:
Reporting should be a function of collecting available context and reporting accordingly. 
- First iteration starts with component-definition and reports controls available. 
- With the presence of a catalog/profile -> we can overlay that control information to calculate percentages. 
- with the presence of assessment-results we can check satisfied/not-satisfied 

Given that - we can start migrating functionality to the oscal model libraries for each model as applicable to get the context we need to return here for collective reporting. 

My  opinion is that reporting is not singly-model focused and rather total context driven. I may be very wrong - let's chat as a team if anyone feels otherwise.

src/cmd/report/report.go
CloudBeard: Swapped that to required removed the check.

src/cmd/report/report.go
CloudBeard: Added more data/examples in the help

src/cmd/report/report.go
CloudBeard: I ended up splitting that up twice, the first time just to separate duties but ended up adding 2 new functions. 1 that is public that will run the actual logic and the second that is a handler for model handlers.

Ran into an issue testing so I came back to add the function that runs the logic.

src/cmd/report/report.go
CloudBeard: I agree but not 100% sure I implemented this part. 

Long term if we run `lula report -f assessment-results.yaml` and we have an OSCAL model for all models then the SAR imports the SAP, SAP imports the SSP, and the SSP imports the Components so we'd have a lot of data to pull from. Ive also thought about locality of the OSCAL.

Short term/this PR im not sure what else to pull out from just the component definitions until we have profile/catalog support. I can see some more there but can be added to. I was thinking new handler per model that can call X functions to do logic work on that model. Also reach across handlers for intertwined models for processing.

Near example would be when we have that profile and catalog support in fully, report on component definition will give you data around the catalog/profile too but use their handlers/functions.

src/cmd/report/report.go
```diff
@@ -0,0 +1,314 @@
...
+
+	spinner = message.NewProgressSpinner("Reading OSCAL model from file")
+	oscalModels, err := readOSCALModel(getOSCALModelsFile)
+	if err != nil {
+		spinner.Fatalf(fmt.Errorf("failed to read OSCAL Model data: %v", err), "failed to read OSCAL Model")
+	}
+	spinner.Success()
+
+	spinner = message.NewProgressSpinner("Determining OSCAL model type")
+	modelType, err := determineOSCALModel(oscalModels)
```
brandtkeller: The `oscal` pkg under `src/pkg/common/oscal` contains a `GetOscalModel()` that returns the same values. 

I would consider removing this -> removing the required parameter from the `handleOSCALModel()` function and call it directly inside `handleOSCALModel()`. 

Unless you have some unique requirement.

src/test/e2e/report_test.go
```diff
@@ -0,0 +1,70 @@
+package test
+
+import (
+    "testing"
+
+    "github.com/defenseunicorns/lula/src/cmd/report"
+)
+
+func TestReportCommand(t *testing.T) {
```
brandtkeller: Any reason these cannot serve as more of a unit test to be coupled with the `report` package. 

I'd like to start iterating `E2E` tests to be more CLI-driven context.

src/test/e2e/report_test.go
```diff
@@ -0,0 +1,70 @@
...
+			expectErr:  false,
+            fileFormat: "json",
+		},
+		{
+			name:       "Component Definition with one sources and one framework props",
+			inputFile:  "../../test/e2e/scenarios/report-data/valid-component-with-framework.yaml",
+			expectErr:  false,
+		},
+        {
+            name:       "Catalog OSCAL Model",
```
brandtkeller: in addition to the conversion of this test to more of a unit test -> it may help reduce testing file requirements. Possible use of data structures and objects if possible (look at your function inputs/outputs and consider the requirements).

src/cmd/report/report.go
```diff
@@ -0,0 +1,314 @@
...
+        return fmt.Errorf("unknown OSCAL model type: %s", modelType)
+    }
+
+    spinner.Success()
+    message.Info(fmt.Sprintf("Successfully processed OSCAL model: %s", modelType))
+    return nil
+}
+
+// Handler for Component Definition OSCAL files to create the report
+func handleComponentDefinition(componentDefinition *oscalTypes_1_1_2.ComponentDefinition, format string) error {
```
brandtkeller: Should we consider composing comp-defs here prior to processing?

brandtkeller (changes requested):
No issues with the current direction. 

I would remove the current documentation and run a `lula internal gen-cli-docs` in order to generate the new CLI docs for this command. Iterate as necessary to try and capture all context in the CLI command code as possible. If you need additional docs place them outside the `docs/cli-commands` directory. 

Otherwise this gets us moving in the right direction - as I mentioned previously, I expect future iterations to be less of a model-by-model reporting structure and rather wholistic reporting of all models - but that is out of scope currently.

src/cmd/report/report.go
brandtkeller: I believe we have a common vision here - but we are blocked by other circumstances so I'm good with pressing in the current direction and evolving this design over time,

src/cmd/report/report.go
CloudBeard: I ended up using a few more functions from common where I found redundancy.

src/test/e2e/report_test.go
CloudBeard: Swapped those to unit and I think created a more CLI approach.

src/test/e2e/report_test.go
CloudBeard: Created a mock component definition data structure to have some data for testing.

src/cmd/report/report.go
CloudBeard: Good call! added the compose pieces to the beginning there.

src/cmd/report/report.go
```diff
@@ -0,0 +1,294 @@
...
+
+To create a new report in yaml format:
+lula report -f oscal-component-definition.yaml --file-format yaml
+`
+
+var reportCmd = &cobra.Command{
+	Use:     "report",
+	Hidden:  false,
+	Aliases: []string{"r"},
+	Short:   "Build a compliance report",
```
brandtkeller: Let's challenge this a bit more -> what function does or can this command serve?

> Report on the current state of compliance
or....

src/cmd/report/report.go
```diff
@@ -0,0 +1,294 @@
...
+lula report -f oscal-component-definition.yaml
+
+To create a new report in json format:
+lula report -f oscal-component-definition.yaml --file-format json
+
+To create a new report in yaml format:
+lula report -f oscal-component-definition.yaml --file-format yaml
+`
+
+var reportCmd = &cobra.Command{
```
brandtkeller: I know this has been up for some time - Can we align the creation of this command to `template` such that the same cobra patterns are followed?

src/cmd/report/report.go
```diff
@@ -0,0 +1,294 @@
+package report
```
brandtkeller, line 1: Formatting appears non-standardized across this file (maybe others). 

I'd run a `go fmt src/cmd/report/report.go` to resolve the formatting and review other modified files. This can also be configured in your development environment whenever "saving".

src/test/e2e/report_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/defenseunicorns/lula/src/cmd"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestLulaReportValidComponent checks that the 'lula report' command works with a valid component definition with multiple components and framework prop.
+func TestLulaReportValidComponent(t *testing.T) {
```
brandtkeller: Heading in the right direction. We have some examples and test utilities to pull from. Would like to standardize on that and @meganwolf0 may have some insights for what we need for structure.

brandtkeller (changes requested):

src/cmd/report/report.go
CloudBeard: Maybe not build a compliance report as in a pdf but I can see it pulling various statics and data out of the OSCAL to be easier to read and understand. I am picturing a thousands of lines SSP and I want to know what controls am I missing, what controls do I have, do I have multiple frameworks and can I see that per framework. For POAMs Im thinking how do I see my critical or highs specifically, how do I see time left on moderates to patch. 

For components I would like to have a quick count of controls per framework, how many have validations, how many controls are in this component-definition that aren't in a profile/catalog. 

I can also see parts of it being reused similar to `lula evaluate -f assessment-result.yaml --sumnmary` doing the summary part. 

Maybe the part that gets the data for UDS Runtime to present too. Thought there is it would be similar questions/data points around the OSCAL to show.

src/cmd/report/report.go
```diff
@@ -0,0 +1,294 @@
...
+
+	return sourceMap, frameworkMap
+}
+
+// Helper to determine if the controlMap source is a URL
+func isURL(str string) bool {
+	return strings.HasPrefix(str, "http://") || strings.HasPrefix(str, "https://")
+}
+
+func PrintReport(data ReportData, format string) error {
```
meganwolf0: Haven't picked through this, so this function might not be relevant, but the `Table` under [messages](https://github.com/defenseunicorns/lula/blob/1ab0eefdeeb1d59f16f33249b1a6fce141ef5942/src/pkg/message/message.go#L322) might save some of this functionality

src/test/e2e/report_test.go
meganwolf0: Looks like this might be in reference to doing some golden file checks, I think the [PR for compose](https://github.com/defenseunicorns/lula/pull/686/files#diff-1a3ee14400dbe38c385f9e2c566a655860814625b79fd3b6dd831a9e2d275c8a) should provide some reusable functions for checking the data from report, and the `tools_compose_test.go` and `tools_template_test.go` might be decent-ish places to start from

meganwolf0: (I think there was also a discussion we were going to have as a team about e2e testing practices/structure so some of that could change)

src/cmd/report/report.go
CloudBeard: Believe ive restructured this

src/cmd/report/report.go
CloudBeard: Just ran this and updated my settings in VScode for futures.

src/cmd/report/report.go
CloudBeard: I added that table functionality, I like the way that looks.

src/test/e2e/report_test.go
CloudBeard: I believe ive followed the same structure

meganwolf0:
Just tested locally - is this `il4` supposed to be down in the framework table?
CleanShot 2024-10-11 at 09 07 58

CloudBeard:
> Just tested locally - is this `il4` supposed to be down in the framework table? CleanShot 2024-10-11 at 09 07 58

Yea that would make sense. Ill fix that real quick-ish

src/internal/reporting/helpers.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/defenseunicorns/lula/src/pkg/message"
+)
+
+// Helper to determine if the controlMap source is a URL
+func isURL(str string) bool {
+	_, err := network.ParseUrl(str)
+	return err == nil
+}
+
+func fetchOrReadFile(source string) ([]byte, error) {
```
brandtkeller: `network.Fetch()` will return file bytes for local files on the filesystem or remote files on network. I don't believe this functionality is needed unless you have an edge case you're solving for.

src/internal/reporting/helpers.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"os"
+
+	oscalTypes_1_1_2 "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-2"
+	"github.com/defenseunicorns/lula/src/pkg/common/network"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+)
+
+// Helper to determine if the controlMap source is a URL
+func isURL(str string) bool {
```
brandtkeller: Given the comment below I imagine this will be removed - but I do think this has some gaps worth addressing if it were to be kept. 

IE the return statement does not indicate a value that aligns to the function intent.

brandtkeller (changes requested):

src/cmd/report/report.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&inputFile, "input-file", "f", "", "Path to an OSCAL file")
+	cmd.Flags().StringVar(&fileFormat, "file-format", "table", "File format of the report")
+	err := cmd.MarkFlagRequired("input-file")
+	if err != nil {
+		message.Fatal(err, "error initializing upgrade command flags")
```
brandtkeller:
```suggestion
		message.Fatal(err, "error initializing report command flags")
```

src/test/e2e/report_test.go
```diff
@@ -0,0 +1,30 @@
...
```
brandtkeller: Is this file still required given `cmd_report_test.go`?

brandtkeller (changes requested):

brandtkeller (changes requested):
I believe you might have removed the wrong test. 

`src/test/e2e/repor_test.go` feels redundant while the test that was in `src/internal/reporting` might still be valid.

brandtkeller (approved):
Appreciative of you running point here and exposing this as an area of importance for users to give us feedback on.

PR #602, opened 2024-08-15
fix(docs): updated namespace doc locations and associated prop namespace url

### LINKED ISSUE #579: [Question] Namespace level in docs
Not so much a feature request as a question/nit on the docs - The `namespace` docs appear at the top-level menu in the Lula docs, wondering if this would make more sense as a subset of OSCAL, since that's the context in which `namespace` is defined?

Just wanted to raise it as a question/possible move, it feels a little out of place to me right now

### Proposed Solution
Move `namespace` directory under `OSCAL`

### Expected Deliverable

Add a constant for the props `namespace` value

Updating props internal to validate/evaluate to point to the new namespace url

generate updates to update a prop ns from the old url to the new

meganwolf0:
## Description

Updated `ns` directory location to be nested in `oscal`

Updated the GetProps and UpdateProps methods to recognize "legacy" urls, if we have to do this again

I _think_ updated all the docs to reference the new `oscal/ns` url

## Related Issue

Fixes #579 

## Type of change

- [X] Other (security config, docs update, etc)

## Checklist before merging


brandtkeller (approved):
Expected workflows against `uds-core` operated as expected.

PR #607, opened 2024-08-18
feat(configuration): add initial support for Viper command initialization

### LINKED ISSUE #596: Viper support - baseline

Viper support for Lula as an entrypoint for future configuration development

- **Given** a config file or environment variable is defined
- **When** a command is executed
- **Then** viper has been initialized for use

This is immediately useful from a CLI runtime configuration perspective. 

### Additional context

Viper can be used by future integrations for configurations that could be used by template processes

### Expected Deliverable

Viper initialization for commands which might use configuration.

brandtkeller:
## Description

Adds initial command support for viper configuration integrated with commands for initializing viper and detecting configuration file and environment variables. 

## Related Issue

Resolves #596

## Type of change

## Checklist before merging


src/cmd/common/viper.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Specify an alternate config file
+	cfgFile := os.Getenv("LULA_CONFIG")
+
+	// Don't forget to read config either from cfgFile or from home directory!
+	if cfgFile != "" {
+		// Use config file from the flag.
+		v.SetConfigFile(cfgFile)
+	} else {
+		// Search config paths in the current directory and $HOME/.zarf.
```
meganwolf0:
```suggestion
		// Search config paths in the current directory and $HOME/.lula.
```

meganwolf0 (commented):
There is a `PersistentPreRun` in many of the commands that override the rootCmd version -> specifically the dev ones, I see the viper config enabled there but due to this it doesn't seem to be reading the config data. Locally I removed PersistentPreRun from each of the dev commands and updated the parent devCmd as follows:
```
var devCmd = &cobra.Command{
	Use:   "dev",
	Short: "Collection of dev commands to make dev life easier",
	PersistentPreRun: func(cmd *cobra.Command, args []string) {
		// Call the parent's PersistentPreRun
		if parentPreRun := cmd.Parent().PersistentPreRun; parentPreRun != nil {
			parentPreRun(cmd.Parent(), args)
		}
		config.SkipLogFile = true
	},
}
```
I think this should solve it - the tools cmds also have this prerun, not sure if we want to update those as well/what the intended config recognition is there

(edit - happy to commit that fix ^^ if you agree, but refrained if there's more discussion needed, I might be missing something as well)

brandtkeller:
@meganwolf0 ended up coming full circle with a slight modification of your proposed solution. 

Optimally I wanted to have the inheritance of the `PersistentPreRun` as much as possible - which does function when you change the child commands to from `PersistentPreRun` to `PreRun`. Problem here then become the sequence of operations -> root would call `SetupCLI()` which logged it's use of a log file before the child/grandchild commands switched the logging file off. 

Given we don't have much complexity baked into our command pre-runs this seems adequate for solving the problem on the two `dev` and `tools` commands. If we ever define the need we can simply move the `PersistentPreRun` down one level and re-think sequence.

src/cmd/dev/common.go
```diff
@@ -21,6 +22,13 @@ var devCmd = &cobra.Command{
 	Use:     "dev",
 	Aliases: []string{"t"},
```
meganwolf0: Might be good to remove or create a different alias (d?)

brandtkeller: great point - got this added now

meganwolf0 (dismissed):

meganwolf0 (approved):

PR #608, opened 2024-08-19
feat(console): initial tui for component-defn read

### LINKED ISSUE #609: Read-Only Component Definition in TUI

Generate TUI that can ingest an OSCAL model file, and allow a user to interact with the component-definition in a read-only experience. The User should be able to:
- Filter by component
- Filter by Framework
- View Remarks and Description for Each control
- See Validations linked to each control 

### Expected Deliverable

Initial TUI with above features along with some associated documentation

meganwolf0:
## Description

Initial TUI Experimentation for read-only operations on component-defn OSCAL model

## Related Issue

Fixes #

## Type of change

## Checklist before merging


meganwolf0:
Looking for a review on the read-only component-definition features, still need to add some docs - any thought as to where those should go?

src/cmd/tui/tui.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/defenseunicorns/lula/src/pkg/common/oscal"
+	"github.com/defenseunicorns/lula/src/pkg/message"
+	"github.com/spf13/cobra"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+var tuiHelp = `
+To view an OSCAL model in the TUI:
+	lula tui /path/to/oscal-component.yaml
```
brandtkeller: Given the `Fatal` message item below when a file is not specified - can we update the command to include the standard `-f / --input-file` flag and mark it as required? Removes the need for that argument check ,improves native cobra flow and should provide more input validation.

brandtkeller (changes requested):
Appreciative of functionality that thinks about the OSCAL as more than just the scope of NIST 800-53.

meganwolf0:
> Removes the need for that argument check ,improves native cobra flow and should provide more input validation.

So I was thinking about this more and wondering if that truly is more "native cobra" - as the Run command expects `args` and this will not run without that argument, so I'm wondering why is it bad to have that check (`len(args)==0`) ? I agree that it's more aligned with our other usage of flags, but also wondering if we should think more about the conventions we've set up for ourselves. If we are constantly having to put -f for a completely mandatory input, is that really desirable? Should flags be more in the realm of optional or intended to have alternative behaviors or additional functionality, as opposed to enforcing mandatory fields (e.g., file inputs)...

Anyway, I can change this, it's more just a thought about why/when to use flags vs args.

src/internal/tui/component/component.go
```diff
@@ -0,0 +1,459 @@
...
+						m.validations.SetItems(make([]blist.Item, 0))
+
+						m.inFrameworkOverlay = false
+					} else {
+						m.inFrameworkOverlay = true
+						m.frameworkPicker.SetContent(m.updateFrameworkPickerContent())
+					}
+
+				case focusControls:
+					m.selectedControl = m.controls.SelectedItem().(control)
```
CloudBeard: Does something like  `if selectedControl := m.selectedControl.SelectedItem().(control) != nil` to check if the thing you're searching for in the control is there to prevent crashes?

CloudBeard: Similar to line 288 is what im basing it on.

CloudBeard:
Finally got hands on with the TUI and its so cool, big fan.

I feel like the -f starts to fall into the log/stdout usage area.

src/internal/tui/component/component.go
meganwolf0: Not sure I'm 100% following, but I believe you don't need to check the filter since that's handled by the bubbles/list model.
This portion
```go
if selectedItem := m.validations.SelectedItem(); selectedItem != nil {
  m.selectedValidation = selectedItem.(validationLink)
}
```
was implemented because you could hit a case where the controls have no validations linked, so those cases the "selectedItem" would be nil.

If you are poking at the case where there's no implemented requirements (e.g., controls) - I just tried importing a file with that as empty and the oscal linting seems to block it... so idk if that's invalid oscal?

CloudBeard: OOO So what I was doing was typing this into the filter on Controls List and I noticed if it wasn't in the control text and I hit enter it would crash

image

image


This dosen't happen in the validations section I think becuase of 

```go
if selectedItem := m.validations.SelectedItem(); selectedItem != nil {
  m.selectedValidation = selectedItem.(validationLink)
}
```

meganwolf0: Ah ok thanks for finding that case - I'll update this!

brandtkeller:
Is there any common testing patterns for this model of TUI? 

Open to making that a separate deliverable but I imagine we will begin hitting many edge cases where a separation of how the model is visualized is separate from unit testing functions.

meganwolf0:
> Is there any common testing patterns for this model of TUI?

hmmm maybe we could look at adding VHS for testing - might need a human to review the gifs but possibly better than nothing? Looks like this [teatest](https://pkg.go.dev/github.com/charmbracelet/x/exp/teatest) library kind of exists, but the fine print on the repo doesn't guarantee any backward compatibility or future support...

brandtkeller:
> > Is there any common testing patterns for this model of TUI?
> 
> hmmm maybe we could look at adding VHS for testing - might need a human to review the gifs but possibly better than nothing? Looks like this [teatest](https://pkg.go.dev/github.com/charmbracelet/x/exp/teatest) library kind of exists, but the fine print on the repo doesn't guarantee any backward compatibility or future support...

I'd lean towards TeaTest given that we manage that dependency accordingly. I'd like to review capabilities (again maybe a separate issue) and annotate gaps -> maybe there is room to contribute back.

meganwolf0:
> I'd lean towards TeaTest given that we manage that dependency accordingly. I'd like to review capabilities (again maybe a separate issue) and annotate gaps -> maybe there is room to contribute back.

Ok that's fair. Is that something you'd like me to try and incorporate in this PR or write another issue (file it under tech debt haha) and tackle later? Just trying to make sure there's nothing remaining in my actions list on this one

brandtkeller:
> > Removes the need for that argument check ,improves native cobra flow and should provide more input validation.
> 
> So I was thinking about this more and wondering if that truly is more "native cobra" - as the Run command expects `args` and this will not run without that argument, so I'm wondering why is it bad to have that check (`len(args)==0`) ? I agree that it's more aligned with our other usage of flags, but also wondering if we should think more about the conventions we've set up for ourselves. If we are constantly having to put -f for a completely mandatory input, is that really desirable? Should flags be more in the realm of optional or intended to have alternative behaviors or additional functionality, as opposed to enforcing mandatory fields (e.g., file inputs)...
> 
> Anyway, I can change this, it's more just a thought about why/when to use flags vs args.

"native cobra" may not be fully capturing intent - but I do want to be iterating towards both flags and args having more strict control using cobra builtins such that we're improving our intended user experience without having to instrument that behavior for each command. 

Things such as combining required flag definitions with different permutations of [argument use](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#positional-and-custom-arguments) such that the pattern improves.

brandtkeller:
> > I'd lean towards TeaTest given that we manage that dependency accordingly. I'd like to review capabilities (again maybe a separate issue) and annotate gaps -> maybe there is room to contribute back.
> 
> Ok that's fair. Is that something you'd like me to try and incorporate in this PR or write another issue (file it under tech debt haha) and tackle later? Just trying to make sure there's nothing remaining in my actions list on this one

New issue to track testing -> would like to gauge interest in getting the testing structure in place prior to additional TUI feature development.

meganwolf0:
For posterity, #627 opened to address testing

docs/console/README.md
```diff
@@ -0,0 +1,8 @@
+# Console
+
+The Lula Console is a text-based terminaluser interface that allows users to interact with the OSCAL documents in a more intuitive and visual way.
```
brandtkeller:
```suggestion
The Lula Console is a text-based terminal user interface that allows users to interact with the OSCAL documents in a more intuitive and visual way.
```

docs/console/README.md
```diff
@@ -0,0 +1,8 @@
+# Console
```
brandtkeller, line 1: Agree that we will want separate documentation from the CLI commands for `console`. Just making sure you intended for this to be a root level docs directory such that it will have it's own header/nav-bar-item on the docs homepage?

brandtkeller (changes requested):
Minor comments but otherwise love the direction this sets. 

 The current structure I believe lends itself well to current thoughts around reporting - where we might be able to pass in multiple OSCAL files and establish an `oscalModel` that contains a valid artifact for each model (maybe this is multiple `-f` flags populating the full `oscalModel` object etc). 
 
 This creates a wholistic context of how the pieces connect to one-another and the ability to execute actions (validate etc) from this view will be fun to pursue.

brandtkeller:
Maybe last item: Can you switch `log_level` to `info` in the `lula-config.yaml` and run:
```bash
./bin/lula internal gen-cli-docs
```

Should only create a doc for the new console command.

docs/console/README.md
meganwolf0: Yeah - I wasn't really sure where to put this to be honest. Root level felt like very aggressively showy, but it also didn't feel like it aligned to any of the other current sections? If you have other thoughts on where this belongs then certainly all ears!

brandtkeller: Let's keep it and push forward

brandtkeller (approved):

Issue #301, opened 2024-03-09
Ability to reproduce any observation 


brandtkeller:
# Intent
For the Lula validation process - we have access to the validations (proofs) of what is measured and the policy it must adhere to - this can be linked and/or transient across OSCAL as the project defines. 

The missing piece for historical assessment data is to have the collected data present in the `Assessment-Results` such that anyone performing an audit of what was assessed would have the ability to "replay" a given point-in-time observation. 

I believe this creates a layer of trust where auditing can be a function of reviewing both the validation inputs, data applied against the inputs, and the policy decision that was made to influence the finding state. 

## Potential Considerations
- encode the data collected from domains that is processed by a provider and include it in the `assessment-result` in some way. 
- Create a flag in the validate command that allows for the replay of a specific observation? Make the output more verbose by logging a pretty-json object and the results of this offline-replay.
- Other functionality?

## Sub-issues
Since this is a decent chunk of work, the following issues have been created to close out this issue:
- Save resources from validation: #620 
- Dev commands to pull resources/validations from assessment results to enable dev replay: #621
- Validate replay: #623

brandtkeller:
Is there other metadata that we need to include in the observation other information that is pertinent for reproducible result.

brandtkeller:
Should we consider the use of a separate file for storing the data collected?

brandtkeller:
This issue has been decomposed into the aforementioned issues. Closing this issue as complete.

PR #612, opened 2024-08-21
feat(validate): save validation resources

### LINKED ISSUE #560: Standardized Output for file write

There are a couple disparate writing processes for Lula that we should consolidate into a common approach.

Compose: 
- Without `-o` flag will write a default `component-composed.yaml` to the same location as the non-composed file
- With `-o` flag will write file to `pwd` as a base for where to write the file

Validate/generate/evaluate/etc:
- Without `-o` flag will write a default `assessment-results` file to `pwd`
- With `-o` flag will write file to `pwd` as a base for where to write the file

Objective is to standardize the process.

### LINKED ISSUE #620: Save resources from validation execution

Sub-task of #301:

In order to reproduce observations, the resources that provided the input to the validation need to be saved, as they are transient across time. 

- **Given** resources would like to be saved from a `lula validate` run
- **When** `--save-resources` is added
- **Then** resources are saved

none

### Additional context

none

### Expected Deliverable

Adding the functionality to "save resources" to both remote files or the back-matter

Updating existing tests + adding new tests as necessary

Update CLI command docs

meganwolf0:
## Description

Adds a flag `save-resources` to `validate` to save the resources to the backmatter of an assessment result

## Related Issue

Relates to #301

## Type of change

## Checklist before merging


brandtkeller (changes requested):
Documenting for posterity:

Moving forward with consideration of the use of back matter more carefully. If we can meet our outcomes with a separate external file as-is today - we should execute on that strategy now while evaluating whether we continue to repeat the back matter storage pattern.

meganwolf0:
Removed back-matter option and associated code changes needed, definitely simplifies things!

src/cmd/validate/validate.go
```diff
@@ -47,9 +47,13 @@ var validateCmd = &cobra.Command{
 	Long:    "Lula Validation of an OSCAL component definition",
 	Example: validateHelp,
 	Run: func(cmd *cobra.Command, componentDefinitionPath []string) {
-		if opts.InputFile == "" {
-			message.Fatal(errors.New("flag input-file is not set"),
-				"Please specify an input file with the -f flag")
+		outputFile := opts.OutputFile
+		if outputFile == "" {
```
brandtkeller, line 51: Are we taking a stance towards standardizing the output of artifacts to be co-located with the targeted input file by default? 

If that was the intent - wanted to ensure we update the issue we have open to reflect a decision being made.

src/cmd/evaluate/evaluate.go
```diff
@@ -134,6 +143,20 @@ func evaluateTarget(target oscal.EvalResult, source string, summary bool) error
 			}
 		}
 
+		// Print machine-readable output
+		if machine {
```
brandtkeller: Can you help me understand what drives the need for this requirement? 

I need to find a good test that will highlight it's value - but I am failing to come up with the intuitive need for `Print a machine-readable output` when that is the point of OSCAL and what we do by default. 

^ that is mainly just the story telling piece - obviously this drills down into observations and may be another helpful flag - I'm just looking to understand the requirement so that we can tell a unified story.

src/pkg/common/oscal/assessment-results.go
```diff
@@ -310,16 +311,34 @@ func FilterResults(resultMap map[string]*oscalTypes_1_1_2.AssessmentResults) map
...
+		UUID:             observationUuid,
 		Description:      fmt.Sprintf(descriptionPattern, descriptionArgs...),
 		RelevantEvidence: relevantEvidence,
 	}
+	// TODO: should the props be added regardless?
+	if resourcesHref != "" {
+		observation.Props = &[]oscalTypes_1_1_2.Property{
+			{
+				Name:  "validation",
+				Ns:    "https://docs.lula.dev/oscal/ns/validation",
```
brandtkeller:
```suggestion
				Ns:    "https://docs.lula.dev/oscal/ns",
```

src/pkg/common/oscal/assessment-results.go
```diff
@@ -310,16 +311,34 @@ func FilterResults(resultMap map[string]*oscalTypes_1_1_2.AssessmentResults) map
...
-		UUID:             uuid,
+		UUID:             observationUuid,
 		Description:      fmt.Sprintf(descriptionPattern, descriptionArgs...),
 		RelevantEvidence: relevantEvidence,
 	}
+	// TODO: should the props be added regardless?
+	if resourcesHref != "" {
+		observation.Props = &[]oscalTypes_1_1_2.Property{
+			{
+				Name:  "validation",
```
brandtkeller, line 329: Need to update docs for this new property

src/pkg/common/oscal/assessment-results.go
```diff
@@ -310,16 +311,34 @@ func FilterResults(resultMap map[string]*oscalTypes_1_1_2.AssessmentResults) map
...
+
+	observation := oscalTypes_1_1_2.Observation{
 		Collected:        rfc3339Time,
 		Methods:          []string{method},
-		UUID:             uuid,
+		UUID:             observationUuid,
 		Description:      fmt.Sprintf(descriptionPattern, descriptionArgs...),
 		RelevantEvidence: relevantEvidence,
 	}
+	// TODO: should the props be added regardless?
```
brandtkeller, line 325: Is this `validation` prop being consumed somewhere?

brandtkeller (changes requested):
Few items to cleanup. Also understand this was a discovery and exploration driven PR as opposed to a direct issue deliverable - so I have no issues with undocumented requirements/additions (in the issue).

src/cmd/validate/validate.go
meganwolf0: I was thinking yeah - since we say that `composed` files are co-located with the input-file location by default - and to me this makes more sense then it being at whatever directory you run the command from. I think there is an issue that kind of covers this, but happy to have more of a discussion around it. I just needed to enforce a location so I could write the resources there as well. And if it was in the working directory I didn't like how the resources would also be added there. 
Are you good with that change? Where should I update stuff if so, I'm not sure any docs right now specify default file location for `assessment-results`, but will need to double check that

src/cmd/evaluate/evaluate.go
meganwolf0: yeah this part is literally just returning a string that I can tell chatgpt to specifically look for and write follow-on commands given the content. I hid this flag deliberately because it makes no sense to an end user, maybe there's a better name though

meganwolf0: we also do a thing where we write to stderr - I could change to writing this machine data to stdout, since it's supposed to be some structured data... Idk that is probably wrapped up in a larger discussion of logging things, I just went with this because it was the easiest to implement as a gap filler

src/pkg/common/oscal/assessment-results.go
meganwolf0: not currently, the [PR](https://github.com/defenseunicorns/lula/pull/622/files#diff-90d2f3923e5be375196f94724668ca04eb49d12ec3cb29c6bd369e8dd54d4ed8) I have in draft (and will need to deconflict with this) has some sample dev print-x commands that use this prop

meganwolf0: Before I update the docs, I think maybe this question stands as something to answer - right now this prop is intended to provide a linkage back to the source validation, we don't actually have any mechanism to do so. Should this be added regardless of if the resources are saved? To me it seems like it would be helpful

src/cmd/validate/validate.go
brandtkeller: I don't have any strong concerns - I can't think of any current use that will be broken.

src/pkg/common/oscal/assessment-results.go
brandtkeller: It provides a traceable pattern that we can refer to in the event we need to modify.

src/cmd/evaluate/evaluate.go
brandtkeller: If you see a requirement of stdout here then that sounds like a good approach.

src/pkg/common/oscal/assessment-results.go
meganwolf0: added docs for this

src/cmd/evaluate/evaluate.go
meganwolf0: Ok, I updated this a bit -
I modified to a json output, since that feels more inline with what should go to stdout. Added a test for the GetMachineFriendlyObservations function.
Now if you run like a out=$(lula evaluate -f assessment-results --machine), the out will just be that json structure

brandtkeller (approved):

PR #644, opened 2024-09-06
feat(template): template command with initial docs

### LINKED ISSUE #591: Template functionality and tools command

Capability to template a targeted file using a (proposed) `lula tools template` command

- **Given** a configuration file
- **When** template is executed
- **Then** variables in the validation are templated into the file

### Expected Deliverable

Functionality to support templating a targeted file given configuration variables and a file with appropriate templates present.

brandtkeller:
## Description

Adds a `lula tools template` command and associated functionality to apply generic templating of configuration data from a configuration file or environment variables.

Structure here is quite loose - allowing for the most permissive optionality for using configuration data (environment variable / config file - with command line to come) such that we can iterate on guard rails as required - otherwise keeping the structure dynamic. 

Adds some new CLI tests under E2E with some more work to be done to generalize and make generic. 

## How to Review

- Review the command and associated logic
- Review the tests - particularly the `src/test/e2e/standard` tests which use testing files available for local testing

## Related Issue

Fixes #591 

## Type of change

## Checklist before merging


comp.yaml
```diff
@@ -0,0 +1,69 @@
+component-definition:
```
meganwolf0: I assume this is a test file?

src/test/unit/common/oscal/valid-component-template.yaml
```diff
@@ -0,0 +1,107 @@
+# add the descriptions inline
```
meganwolf0, line 1: this file is failing the `lint` command - is this ok? I suppose I was thinking these template files would still be valid oscal (maybe we just need to encapsulate the {{ .. }} in ""?

src/internal/template/template_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	t.Run("Test {{ .testVar }} but empty data", func(t *testing.T) {
+		data := map[string]interface{}{}
+
+		test(t, data, "{{ .testVar }}", "<no value>")
+	})
+
+}
+
+func TestGetEnvVars(t *testing.T) {
```
meganwolf0, line 42: It look like if I do "LULA_TYPE" and template the sample file, the type from the config file gets overridden, which is what I'd expect. I'm not sure if/how I'd override nested keys (e.g., resources.secret). Also should there be a test for testing the envvar templates/overrides?

meganwolf0 (changes requested):

comp.yaml
brandtkeller: yes!

src/internal/template/template_test.go
brandtkeller: Agree on the nested override option via environment variables - this highlights a constraint that we should review as environment variables may be simple string replacements until we define a process. I don't see that definition as potentially non-backwards compatible.

I like the idea of testing preceence for overrides. I could do so easily in the "E2E" tests if we wanted to do some testing there. Open to thoughts for introducing that here.

src/test/unit/common/oscal/valid-component-template.yaml
brandtkeller: Part of my running in auto-pilot here. I currently see a future-documented stance where we expect linting to pass and other scenarios where we do not. The `template` command would be the more permissive stance - where other commands might expect valid yaml/json as a constraint.

I could go either way here or maybe following PR's changes that stance entirely. 

I will go ahead and make this one valid.

src/internal/template/template_test.go
brandtkeller: Added a quick override with an environment variable to the `E2E` test that changes the expectation for an item with configuration also present in the config file.

meganwolf0 (approved):

PR #646, opened 2024-09-07
fix(gen-cli-docs): fix path for gen-cli-docs

### LINKED ISSUE #645: Broken links in cli command docs
### Steps to reproduce
1. visit a command containing subcommands [example](https://docs.lula.dev/cli-commands/lula_dev/)
2. click on one of the subcommand links
  
### Expected result

Route to the subcommand documentations

### Actual Result

404

### Severity/Priority

Medium

### Additional Context

Need to update the `gen-cli-command`

brandtkeller:
## Description

Update the `gen-cli-docs` command to generate links that the docs website will resolve. The reference needs to originate from root ie `/cli-commands/<command>`

## Related Issue

Fixes #645 

## Type of change

## Checklist before merging

### DIFF
--- a/docs/cli-commands/lula.md
+++ b/docs/cli-commands/lula.md
@@ -20,12 +20,12 @@ Real Time Risk Transparency through automated validation
 
 ### SEE ALSO
 
-* [lula completion](/cli-commands/lula_completion/)	 - Generate the autocompletion script for the specified shell
-* [lula console](/cli-commands/lula_console/)	 - Console terminal user interface for OSCAL models
-* [lula dev](/cli-commands/lula_dev/)	 - Collection of dev commands to make dev life easier
-* [lula evaluate](/cli-commands/lula_evaluate/)	 - evaluate two results of a Security Assessment Results
-* [lula generate](/cli-commands/lula_generate/)	 - Generate a specified compliance artifact template
-* [lula tools](/cli-commands/lula_tools/)	 - Collection of additional commands to make OSCAL easier
-* [lula validate](/cli-commands/lula_validate/)	 - validate an OSCAL component definition
-* [lula version](/cli-commands/lula_version/)	 - Shows the current version of the Lula binary
+* [lula completion](./lula_completion.md)	 - Generate the autocompletion script for the specified shell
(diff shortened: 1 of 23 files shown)


meganwolf0:
are these supposed to be links to the `*.md` documents? (I think the urls would just be the name of the file, no?)

brandtkeller:
> are these supposed to be links to the `*.md` documents? (I think the urls would just be the name of the file, no?)

These are links to the resolvable page in the docs website minus the base URL `docs.lula.dev`

Example: 

`docs.lula.dev` + `/cli-commands/lula_dev/` would route to the page accordingly. 

the current `/cli` was an oversight and that obviously (in hindsight) doesn't route properly.

meganwolf0:
If you just do like ./name.md though wouldn't that be handled by the Lula-docs code and also be a usable link in the repo?

brandtkeller:
> If you just do like ./name.md though wouldn't that be handled by the Lula-docs code and also be a usable link in the repo?

It could be - this is where more testing with lula-docs interpretation is required. I was simply noting the delta between current link and routing and fixing accordingly.

meganwolf0:
> It could be - this is where more testing with lula-docs interpretation is required. I was simply noting the delta between current link and routing and fixing accordingly.

If you see the README.md in `getting-started` it has a relative link to `./simple-demo.md` which is properly handled by the [website](https://docs.lula.dev/getting-started/#tutorials). So I guess this is really a matter of if we want the links to work both in the lula repo and on the site - if yes then maybe we update to just point to the file name and if the logic is consistent it should remap to the site correctly

brandtkeller:
> If you see the README.md in `getting-started` it has a relative link to `./simple-demo.md` which is properly handled by the [website](https://docs.lula.dev/getting-started/#tutorials). So I guess this is really a matter of if we want the links to work both in the lula repo and on the site - if yes then maybe we update to just point to the file name and if the logic is consistent it should remap to the site correctly

Great point - provided the docs infrastructure (hugo) does allow those relative links (as you've indicated) then updating to support both is pretty trivial given the flat structure. Updated accordingly.

meganwolf0 (approved):

PR #667, opened 2024-09-16
feat(validation)!: validation testing framework

### LINKED ISSUE #460: Validation Review Fields/Process
Capture processes and enhancements to the validation process for how a validation is/should be tested. 

- **Given** a decomposed validation exists
- **When** dev validate is executed
- **Then** context and processes are present to test pass/fail scenarios

Perform this as an isolated process from Lula

### Additional context
This could be additional fields in the validation that support how to test the validation. 

Should we trim these off when composing?

meganwolf0:
## Description

Initial work to specify Lula-native validation testing
* Updates the previously (imo poorly) named `inject` function to `transform` - _tried_ to do a bit of refactor to pull out functionality to support more granular tests
* Addition of `RunTests` method to the `LulaValidation` + updated schema to support
* Updated docs -> Hopefully the syntax/intended usage isn't too confusing. Added some sample tests for existing validations we have in compliance artifacts (https://github.com/defenseunicorns/compliance-artifacts/pull/151) - Will note that the one functionality not currently supported is deleting a specific list entry. Was thinking this could be a follow-on because this is already quite lengthy
* Modified the `dev validate` to handle `--run-tests` flag, and because this was getting annoying to debug a `--print-test-resources` that will print out the transformed resources.json to the validation directory -> This is basically just a print-to-screen implementation right now. Probably further discussion needed on dumping a test-result object or how we want to handle that.

Follow-ups to this
* probably a refactor of the `dev validate` cmd to support the e2e testing structure we have
* addition of `--run-tests` to `lula validate` -> will need to include some exploration/decision on how the test report gets dumped
* delete a specified list item

## Related Issue

Fixes #460 

## Type of change

## Checklist before merging


docs/reference/testing.md
```diff
@@ -0,0 +1,144 @@
+# Testing
+
+Testing is a key part of Lula Validation development. Since the results of the Lula Validations are determined by the policy set by the `provider`, those policies must be tested to ensure they are working as expected.
+
+## Validation Testing
+
+In the Lula Validation, a `tests` property is used to specify the each test that should be performed against the validation. Each test is a map of the following properties:
```
brandtkeller:
```suggestion
In the Lula Validation, a `tests` property is used to specify each test that should be performed against the validation. Each test is a map of the following properties:
```

docs/reference/testing.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `name`: The name of the test
+- `changes`: An array of changes or transformations to be applied to the resources used in the test validation
+- `expected-result`: The expected result of the test - satisfied or not-satisfied
```
brandtkeller, line 11: > The expected result of the test - satisfied or not-satisfied

This is an interesting point as we look to clear a path for validations used outside of OSCAL. We map the result of a policy (pass/fail) to align to findings/observations using the OSCAL terminology (satisfied/not-satisfied). 

We might look at what terminology OPA/Kyverno use and/or decide on what Lulas will be.

brandtkeller (commented):
Minor comments as I look at general structure otherwise no major concerns.

docs/reference/testing.md
meganwolf0: That's fair - I initially was thinking like pass/fail - that also sort of conflates with the overall test result (which is pass/fail) so walked back the terminology. I mean it could be more like policy/provider-result -> true/false? Or accept/reject?

meganwolf0:
Made a couple updates to incorporate the templating for dev commands work
- Added tests for the `dev validate --run-tests` in e2e -> Changed the logic to fail the command when --run-tests is passed and a test fails
- Change the ctx from context.Background() to cmd.Context() - I was thinking that's the context we use in `lula validate` so just being consistent

brandtkeller (commented):
nearly finished with review - no major modifications noticed as of yet.

docs/reference/testing.md
brandtkeller: Walk me through the expected result <> test result of pass/fail and how that might conflate? 

Seems like pass/fail at the test result layer would align well to the expected result also being pass/fail

meganwolf0: The expected-result is like if the policy returns true/false essentially. So I was thinking if expected-result is "fail" (mapping to a policy return of for example, opa: validate.validate=false) then that would "fail" successfully - and the test would "pass". So it just sets up those words to be used together to basically result in opposite outcomes.

This feels like it could be a larger team convo because this is very much a "in my head this is true" sorta thing but I think majority should rule here. I.E., If no one else is confused by a "fail" = "pass" then it's a me problem 😄

brandtkeller: I think that is a fair evaluatino of the problem and good context to have.

brandtkeller (approved):
Developer Experience around the validation authoring process will be super important to influence. Appreciate where you took this.

PR #672, opened 2024-09-18
feat(template)!: introducing variables and sensitive configuration

### LINKED ISSUE #641: Define sensitive configuration and prevent templating

Any operation that templates a file and writes a new file or logs the operation should prevent templating the sensitive templates while retaining the templates for future utilization. 

- **Given** a template is defined as sensitive
- **When** any write operation occurs
- **Then** sensitive values are not templated

N/A

### Additional context

A `template` command may have a flag that explicitly allows for templating the sensitive configuration in order to test validity. 

### Expected Deliverable

Support for handling sensitive templates and configuration data

meganwolf0:
## Description

Sample implementation for sensitive and non-sensitive configuration data to support templating

## Related Issue

Relates to #641

## Type of change

## Checklist before merging


src/internal/template/template.go
```diff
@@ -2,48 +2,188 @@ package template
...
+		},
+		"mask": func(a string) string {
+			return "********"
+		},
+		// Add more custom functions as needed
+	}
+
+	// Parse the template and apply the function map
+	tpl := template.New("template").Funcs(funcMap)
+	tpl.Option("missingkey=zero")
```
brandtkeller: Although not entirely intuitive - I believe this will error under some known scenarios due to the use of a nested key in `constants` and how `missingkey=zero` works. I believe you'd want `missingkey=default` instead.

To reproduce:
```
./bin/lula tools template -f src/test/e2e/standard/validation.tpl --sensitive
```

Note the unexpected error instead of the `the zero value for the map type's element`

src/cmd/tools/template.go
```diff
@@ -42,17 +57,54 @@ var templateCmd = &cobra.Command{
...
+
+		// Handles merging viper config file data + environment variables
+		templateData := template.CollectTemplatingData(constants, variables)
+
+		// override anything that's --set
+		overrideTemplateValues(templateData, templateOpts.Set)
+
+		// Execute the template function based on the flags (TESTING ONLY)
+		var templatedData []byte
+		if templateOpts.Const {
```
brandtkeller: We may want some error handling here for the inclusion of multiple flags which are not currently supported in a single execution. 

IE if I run `lula tools template -f file.yaml --const --all`

brandtkeller (changes requested):
I am a fan of this direction - called out a few items we may want to think about but otherwise this meets our requirements. 

I looked for any concerns or risks present with the nested brace replacement and did not find any issues. 

A lot of commonality between our approaches made this an easy review.

brandtkeller:
Additionally can we update the go mod/sum files when preparing for next review?

src/internal/template/template.go
meganwolf0: I think per the conversation in slack, modified to missingkey=error to try and expose errors as early as possible

src/cmd/tools/template.go
meganwolf0: yeah, this was a dumb testing thing - I update this so you're just passing a single render type

meganwolf0:
> Additionally can we update the go mod/sum files when preparing for next review?

When you say update - is there a particular package you were looking at or just like a `go mod tidy` cleanup?

meganwolf0:
So I spent a little more time on this then I was initially planning and did some refactor of the internal/template package + the command. I was running into some issues when testing the cmd where the flags getting set were persisting across runs - which is not ideal. I refactored this so that we don't have external to the command flags (which sort of runs counter to what we do in all our other cmds, but I've seen this pattern in other cobra cli usage and tbh I like it better, seems to reduce side effects when testing.)

src/cmd/tools/template.go
```diff
@@ -11,79 +13,115 @@ import (
...
+	cmd.MarkFlagRequired("input-file")
+	cmd.Flags().StringVarP(&outputFile, "output-file", "o", "", "the path to the output file. If not specified, the output file will be directed to stdout")
+	cmd.Flags().StringSliceVarP(&setOpts, "set", "s", []string{}, "set a value in the template data")
+	cmd.Flags().StringVarP(&renderTypeString, "render", "r", "masked", "values to render the template with, options are: masked, constants, non-sensitive, all")
+
+	return cmd
 }
 
 func init() {
 	common.InitViper()
```
meganwolf0, line 114: Also - is this redundant with the InitViper called at root? We have this in all the commands and I don't think I gave it much thought before but digging into the viper package we have I don't think we actually need to invoke this?

src/cmd/tools/template.go
```diff
@@ -11,79 +13,115 @@ import (
...
-To template an OSCAL Model:
+To template an OSCAL Model, defaults to masking sensitive variables:
 	lula tools template -f ./oscal-component.yaml
 
 To indicate a specific output file:
 	lula tools template -f ./oscal-component.yaml -o templated-oscal-component.yaml
 
-Data for the templating should be stored under the 'variables' configuration item in a lula-config.yaml file
+To perform overrides on the template data:
+	lula tools template -f ./oscal-component.yaml --set var.key1=value1 --set const.key2=value2
```
brandtkeller:
```suggestion
	lula tools template -f ./oscal-component.yaml --set .var.key1=value1 --set .const.key2=value2
```

src/cmd/tools/template.go
brandtkeller: I believe you are correct - I can't find a permutation where this needs to be executed here. 

Let's adjust this in a separate issue/pr such that we can trace origins of the change in the event it does break something. 

(I'd also like to include some scope in that issue which documents how `init()` is meant to be used such that we have evidence to support the change.

brandtkeller (changes requested):
Super minor inconsistency in the help command. quick modification and re-generation of the cli doc and I'm good!

brandtkeller (approved):

src/cmd/tools/template.go
brandtkeller: Thinking on this more - it's certainly not explicitly required for commands - but it is used on a per-command basis when we perform lookups of viper configurations during the `init()` process. 

(Getting the content of an environment variable that Viper has automatically registered requires `GetString()` or applicable operation)

PR #686, opened 2024-09-27
feat(compose): template files during compose operations

### LINKED ISSUE #642: Template files during persistent writes

Ability to perform templating during compose or retain original templates 

- **Given** templates in an oscal file
- **When** compose is executed
- **Then** optionality to either template the non-sensitive fields or retain templates during file write

### Additional context

We will likely want the ability to support both

### Expected Deliverable

- [ ] Compose supports retaining templates (default)
- [ ] Compose supports performing templating with a flag

meganwolf0:
## Description

Sorry for another massive PR - just to hopefully make the review a little easier going to try and enumerate the bigger things that changed here -

- `src/cmd/common/viper.go` -> made TemplateConstants and TemplateVariables global. Don't love this pattern but in the interest of not totally redoing this and not having to call "GetTemplateConfig" in every template-able workflow, just ideally will run this once in `InitViper`
- `src/cmd/tools/compose.go` -> refactored this command a bit to enable just calling the cmd (mostly for testing, but also I feel like returning the command is a better pattern?)
- `src/pkg/common/composition` -> added the CompositionContext so it would more easily pass the TemplateRenderer, which basically get used in the sub functions to do the template rendering pieces. Includes `options` to set those values, I am a fan of this pattern but may not have really done this the best way so if some of this is an anti-pattern lmk
- `src/pkg/common/network` -> network fetch options to allow for a set base directory, instead of doing the cwd changes. This is pretty coupled to the composition implementation but could possibly be split to a separate PR, it just wouldn't fix the UDS issue on it's own

## Related Issue

Fixes #642 

## Type of change

## Checklist before merging


meganwolf0:
Noticed when testing that case-sensitive keys in `constants` don't turn into case-sensitive keys in the map[string]interface{} that's created from that key (viper thing) - added a note in the configuration.md and filed associated [issue](https://github.com/defenseunicorns/lula/issues/695)

src/pkg/common/composition/composition.go
```diff
@@ -2,44 +2,65 @@ package composition
...
 	"github.com/defenseunicorns/lula/src/pkg/common"
 	"github.com/defenseunicorns/lula/src/pkg/common/network"
 	"github.com/defenseunicorns/lula/src/pkg/common/oscal"
 	"github.com/defenseunicorns/lula/src/pkg/message"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
+type RenderedContent string
+
+type CompositionContext struct {
```
mildwonkey, line 23: I like the pattern!

This is nitpicky, and I am not going to fight for a change here, just want to see what you think. Calling this a context when it isn't actually a `context.Context` is unintuitive to me, especially since you're also using context in the  `ComposeComponentDefinitions ` method. When I see something called a context I'd expect to get values from it using for e.g. `myCompositionCtx.Value(common.LulaWorkDir)`.  What do you think?

meganwolf0: yeah that's totally fair - I guess I was thinking `context` like in that it contains the information needed to "compose" a component definition, but yeah if that's not intuitive/you have another word that could make sense there totally happy to change it!

src/cmd/tools/compose.go
```diff
@@ -14,73 +13,99 @@ import (
...
+		Short:   "compose an OSCAL component definition",
+		Long:    composeLong,
+		Example: composeHelp,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			composeSpinner := message.NewProgressSpinner("Composing %s", inputFile)
+			defer composeSpinner.Stop()
+
+			// Update input/output paths
+			if filepath.IsLocal(inputFile) {
+				inputFile = filepath.Join(filepath.Dir(inputFile), filepath.Base(inputFile))
```
mildwonkey: I'm not sure what this is meant to do - `filepath.Dir()` gets the directory from the input, stripping out the file name, it does not calculate the absolute path. As written, this results in the same thing as the inputFile (same as the outputFile below)

meganwolf0: Ah yep good call - I kept mixing up Dir and Abs - I'll fix it!

meganwolf0: on second thought, just decided to remove those - I don't actually think I need them (the only place I care about the absolute path is for the input file, which I'm calculating in the composition options)

src/cmd/tools/compose.go
```diff
@@ -14,73 +13,92 @@ import (
...
+			defer composeSpinner.Stop()
+
+			if outputFile == "" {
+				outputFile = GetDefaultOutputFile(inputFile)
+			}
+
+			// Check if output file contains a valid OSCAL model
+			_, err = oscal.ValidOSCALModelAtPath(outputFile)
+			if err != nil {
+				message.Fatalf(err, "Output file %s is not a valid OSCAL model: %v", outputFile, err)
```
brandtkeller: Should this be returning an error instead of executing the fatal function?

brandtkeller (commented):
Partial review - getting a baseline understanding and a super minor fix.

src/cmd/tools/compose.go
meganwolf0: yep! nice catch

src/pkg/common/composition/resource-store.go
```diff
@@ -112,35 +117,60 @@ func (s *ResourceStore) AddFromLink(link *oscalTypes_1_1_2.Link) (ids []string,
...
 			ids = append(ids, resource.UUID)
 		}
 	}
 
+	s.SetHrefIds(link.Href, ids)
+
 	return ids, err
 }
+
+func createTemplateResource(data []byte) *oscalTypes_1_1_2.Resource {
```
brandtkeller: Is this unused now that we had discussed not using `props` until we have solid evidence to support when and where to use them?

brandtkeller (changes requested):
Review cont - Can we run `lula internal gen-cli-docs` to update docs?

src/pkg/common/composition/resource-store.go
meganwolf0: yep, removed!

brandtkeller (approved):

src/pkg/common/composition/composition.go
brandtkeller: Given the scope for which this is used - we can continue to iterate on naming if we find a better title.

PR #687, opened 2024-09-27
fix: error checking prior to file writes

### LINKED ISSUE #678: Error checking prior to file writes
### Steps to reproduce

Need to determine steps to reproduce

### Expected result

If some error has occurred that resulted in an OSCAL model not being properly available for write to a file - we should error and/or handle that write accordingly. 

### Actual Result

Occasional errors in a workflow might produce a properly named file with empty contents. This produces additional errors on follow-on write operations that want to perform a merge operation. 

### Severity/Priority

Low - likely edge case driven 

### Additional Context

N/A

meganwolf0:
## Description

Adds functionality to validate if the desired output files for `validate` and `compose` are valid OSCAL models. Also a few additional error detail.

## Related Issue

Fixes #678 

## Type of change

## Checklist before merging


brandtkeller (commented):
I'm looking at some slight restructuring of the OSCAL `pkg` as is - Nonetheless I believe your adjustment to `validate` nails the intent of the (current) discussion in the issue -> checking output validity before executing any processing logic. 

Do we want to map this behavior to all locations that would merge processed data with an existing file?

meganwolf0:
> Do we want to map this behavior to all locations that would merge processed data with an existing file?

Sure, I think the only one that would be missing is `generate`? Did I miss something else?

brandtkeller:
> > Do we want to map this behavior to all locations that would merge processed data with an existing file?
> 
> Sure, I think the only one that would be missing is `generate`? Did I miss something else?

Thinking about the paths now - otherwise I think the only distinction I would make is repeating the pattern of checking for validity of the output file before processing - so scenarios like `evaluate` might need checking up-front as well.

meganwolf0:
> so scenarios like `evaluate` might need checking up-front as well.

It looks like evaluate sort of already has this built in - the first thing it does is grab all the files passed in and convert them to assessment-results, which should basically act as a validation of the output (since the output file is the input file)?

brandtkeller (approved):
Talked through how `evaluate` actively performs this functionality to-date. Covers known paths where a merge operation is expected and handles the check before performing any processing cycles

PR #688, opened 2024-09-27
feat(domains): file domain

### LINKED ISSUE #337: Support for a file domain
Generic file domain for processing a given target could open potential integrations with a variety of tools and reporting.

Notably CVE scanning tools such as Grype & Trivy provide json reporting output formats that Lula could consume and map to applicable controls / other OSCAL artifacts (Such as POAM). In-toto attestations are another example

### Additional Context

Optimally this will be an area that allows for experimentation and testing prior to promoting optimizations of distinct scenarios into new domains.

Scope here should be a single file type (json) - but should support adding additional file type support.

### Issue Deliverables

- [ ] Creation of a `files` domain
- [ ] Ability to specify a file path (local or remote) in a validation
- [ ] Ability to process a `json` file (otherwise error)
- [ ] Ability for each provider to accurately evaluate the data
  - [ ] OPA
  - [ ] Kyverno
- [ ] Docs
- [ ] Tests

### Stretch Goal

- [ ] Add support for yaml files (otherwise create an issue for yaml support)

mildwonkey:
## Description

This adds a `files` domain which use's OPA's `conftest` parser to read in a long list of configuration file types * so that rego validations can be written against the contents. 

```yaml
domain:
  type: file
  file-spec:
    filepaths:
    - name: grafna-config
       path: badconfig.ini
```

I'm not thrilled with the UX right now and I'm not sure what I want to do about it, so here's a draft PR so we can discuss! The file domain takes an array of files, but the implications are that the output `DomainResources` map uses the file name as a key, so you need to include the filename in the validation:

```rego
      package validate
      import rego.v1

      # Default values
      default validate := false
      default msg := "Not evaluated"
      
      validate if {
       check_grafana_config.result
      }
      msg = check_grafana_config.msg
      
      check_grafana_config = {"result": true, "msg": msg} if {
        config := input["badconfig.ini"]
        protocol := config.server.protocol
        protocol == "https"
        msg := "Server protocol is set to https"
      } else = {"result": false, "msg": msg} if {
        config := input["badconfig.ini"]
        protocol := config.server.protocol
        protocol == "http"
        msg := "Protocol is http. That's bad, mmkay?"
      }
``` 

I can make a nice list of arguments why this is good, and why this is bad: on the one hand, we don't risk collision if multiple files have the same field. On the other hand, maybe it's ok if the `files` domain can only read a single file at a time and then we can leave the filename out of the map, or if we choose to merge all the data into a single "file" first. Both feel like big footguns to me. In the former case - it's entirely valid to have for e.g. a dozen terraform files where one will do. you still want to validate it all at once. in the latter - who's responsibility is it to ensure we aren't losing data due to collisions?.

Also this still needs work, tests, and documentation.

From https://www.conftest.dev/:
```
As of today Conftest supports:

CUE
CycloneDX
Dockerfile
EDN
Environment files (.env)
HCL and HCL2
HOCON
Ignore files (.gitignore, .dockerignore)
INI
JSON
Jsonnet
Property files (.properties)
SPDX
TextProto (Protocol Buffers)
TOML
VCL
XML
YAML
```
## Related Issue
Relates to # https://github.com/defenseunicorns/lula/issues/337

- [x] Creation of a files domain
- [X] Ability to specify a file path (local or remote) in a validation
  - [x] local
  - [X] remote
- [x] Ability to process a json file (otherwise error)
- [X] Ability for each provider to accurately evaluate the data
  - [x] OPA
  - [X] Kyverno

## Type of change

## Checklist before merging


brandtkeller (commented):
I appreciate the direction this is driving and the conciseness of the current implementation. 

Your identified constraints for what we do/do-not allow are certainly worth working through. I believe we're at a phase where constraints are valid if documented such that we can point to them - but would love to continue to weigh approaches. 

When it comes to testing - I do want to call out that this should work with any provider - the conftest/opa overlay is obviously very opa heavy (as is much of our implementation with Lula - as OPA/rego is quite powerful) but we do want to ensure we're also testing against the Kyverno baseline. Something we need to practice more often.

mildwonkey:
@brandtkeller yep! I am VERY CONCERNED about how this works with kyverno - honestly I am a bit surprised by how separate the domains and providers are, because I can see a world where certain providers can only support certain domains, or where the domain resources need to be gathered by the provider  so we're certain we're getting a provider-compliant output. That concerns me!

src/pkg/domains/files/files.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// TODO? this might be a nice configurable option (for debugging) - or we go
+	// the terraform route (sorry) and download and store them into a local
+	// .lula directory
+	defer os.RemoveAll(dst)
+
+	// Copy file to a temporarly location, using go-getter to pull any remote files
+	// TODO: use a real context, this isn't correct.
+	g := getter.DefaultClient
```
meganwolf0: We have a `src/common/pkg/network` package (specifically the `Fetch` method) that I think we could also use here

src/pkg/domains/files/files.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return drs, nil
+}
+
+// IsExecutable returns false; the file domain is read-only.
+//
+// This is a tiny lie: the files domain will download remote files into a
+// temporary directory if the file paths are remote, but that is temporary and
+// it is not mutating existing resources so we're calling it non-executable.
+func (d Domain) IsExecutable() bool { return false }
```
meganwolf0, line 89: fwiw, I agree with the non-executable stance - `lula validate` will write a file anyway, so feels like writing tmp files locally is within the purview of the accesses provided to Lula

src/pkg/domains/files/files_spec.go
```diff
@@ -0,0 +1,5 @@
+package files
+
+type Spec struct {
+	Filepaths []string
```
meganwolf0: I think it would be good to allow for naming here - that's sort of our convention with the rest of the domains and that's what would be keyed on in the `provider`. If you have like a long ugly filename it might be kind of annoying to have to key on that down in the provider section. Also I think just makes it slightly easier to swap if file names/locations change

src/pkg/domains/files/files.go
meganwolf0: I think this is also going to struggle when we put it into an actual `validate` workflow - I think we might have to do some passing with the cwd to handle any kind of relative paths because I believe if you execute from not the cwd the relative path gets bonked (I did some updates on `Fetch` to handle some of the wonky file dir changes [here](https://github.com/defenseunicorns/lula/pull/686/files#diff-94793c325bcfb4197d156903e7feac6b8482ec11140af38e1333ff8e132c2d19))

meganwolf0 (changes requested):

meganwolf0:
> we do want to ensure we're also testing against the Kyverno baseline

Does kyverno polices support any data schema or just Kubernetes resources? I should probably know the answer but a quick parsing of their docs seems to indicate it's highly coupled to that format... if so I don't know how we'd use it here

src/pkg/domains/files/files.go
mildwonkey: I like go-getter better for this because it can pull from a longer list of sources, and it pulls directories, and I envisioned pulling things like entire git repos, s3 buckets, and terraform modules into it. If anything I'd replace recommend replacing `Fetch` with go-getter for that reason.

I'm already handling the relative path changing, that's not been an issue in validate, unless I'm misunderstanding something; I'll keep an eye out for issues there.

src/pkg/domains/files/files.go
mildwonkey: thanks! I cleaned up the comment based on feedback <3

src/pkg/domains/files/files_spec.go
mildwonkey: 👍🏻 done!

src/pkg/domains/files/files.go
mildwonkey: 👍🏻  I replaced go-getter w/ network.Fetch for. consistency (for now!)

mildwonkey:
> > we do want to ensure we're also testing against the Kyverno baseline
> 
> Does kyverno polices support any data schema or just Kubernetes resources? I should probably know the answer but a quick parsing of their docs seems to indicate it's highly coupled to that format... if so I don't know how we'd use it here

[`kyverno-json`](https://github.com/kyverno/kyverno-json) can be used with any arbitrary json-formatted data, so it does work reasonably well with the output of `DomainResources`. The UX is icky though, IMO, but I barely know kyverno and OPA so I'm hoping I come to like it with time (and of course as far as I know I'm doing this the clunkiest way possible):

```yaml
provider:
  type: kyverno
  kyverno-spec:
    policy:
      apiVersion: json.kyverno.io/v1alpha1
      kind: ValidatingPolicy
      metadata:
        name: grafana-config
      spec:
        rules:
        - name: protocol-is-https
          assert:
            all:
            - check:
                goodconfig.ini:
                  server:
                    protocol: https
```

docs/reference/domains/file-domain.md
```diff
@@ -0,0 +1,75 @@
...
+* Jsonnet
+* Property files (.properties)
+* SPDX
+* TextProto (Protocol Buffers)
+* TOML
+* VCL
+* XML
+* YAML
+
+## Validations
```
brandtkeller, line 39: We don't yet have a good pattern for the mapping of domain -> provider. 

That said - I do like the Kyverno example below. Could we add an OPA provider example as well?

src/pkg/domains/files/files.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// conftest's parser returns a map[string]interface where the filenames are
+	// the primary map keys.
+	config, err := parser.ParseConfigurations(files)
+	if err != nil {
+		return nil, err
+	}
+
+	// clean up the resources so it's using the filepath.Name as the map key,
+	// istead of the file path.src/pkg/domains/files/files.go
```
brandtkeller:
```suggestion
	// instead of the file path.src/pkg/domains/files/files.go
```

brandtkeller (changes requested):
Appreciative of the code comments and not withstanding the mention of E2E tests am quite excited about this capability.

src/pkg/domains/files/files.go
mildwonkey: I fixed the rest of the line too thank you!

docs/reference/domains/file-domain.md
mildwonkey: I don't think the writing is very good but I added an OPA example! please nitpick freely.

mildwonkey:
I'm still working on this, but I'm going to split it up into a couple of smaller PRs (probably, either way I'm cleaning this mess up)(messing with directories is always a fun time)

Issue #563, opened 2024-07-26
Generate OSCAL Component Source Flag Instead of Catalog and Profile


CloudBeard:
### Is your feature request related to a problem? Please describe.

Component Definitions can be built using a `catalog` or `profile` see [`component-definition`](https://pages.nist.gov/OSCAL-Reference/models/v1.1.2/component-definition/json-reference/#/component-definition/components/control-implementations/source) but the `lula generate component` command requires the `-c` flag and accepts either a catalog or profile while failing if you only provide the `-p` for the profile.

`lula generate component -c https://raw.githubusercontent.com/GSA/fedramp-automation/93ca0e20ff5e54fc04140613476fba80f08e3c7d/dist/content/rev5/baselines/json/FedRAMP_rev5_HIGH-baseline-resolved-profile_catalog.json -r ac-1`

### Describe the solution you'd like

Remove the `-c` and `-p` flags in favor of an `-s` flag for `--source` that accepts either the `catalog` or `profile`.
(rename `-c` to `-s` flags and leave the functionality as it since it accepts both and works.)

### Additional context

![image](https://github.com/user-attachments/assets/af088767-f5a8-4bab-ad55-3f8f85fb1488)

brandtkeller:
Certainly valid issues to call out - one being that the `profile` flag is not currently implemented. 

Second to that - i'd like to think about what the matrix of potential opportunities is - IE generation with:
- both catalog and profile
- no catalog but profile
- catalog but no profile

If the above are all valid - maybe separation of the two could still be warranted?

CloudBeard:
The catalog and profile both go in the same field and is a 1 to 1 was my original thought process.

Now I could see something along the lines of Impact Levels or another reason to map a profile of High/Mod/Low and the catalog because X control is outside of the profile but needs to be represented so needing to do a `control-implementations` for a catalog and a profile could be a thing.

brandtkeller:
revisiting this topic - Profile generation #694 implements a `--source / -s` flag for the use of either profile or catalog to align to this thought. 

Will require some rewiring but I do agree with your original thought and aligning to a 1:1 mapping.

PR #694, opened 2024-10-01
feat(generate): support for profile model and basic generation

### LINKED ISSUE #67: Profile Model Support
Enable option in the CLI to accept which profile is being evaluated in an OSCAL file.

brandtkeller:
## Description

Adds initial support for the profile model with some very basic generation functionality to build upon. 

Looking to implement an OSCAL model interface for two reasons:
1. validate the necessity of the interface across models
2. experiment now vs future alignment with the underlying library possibly changing

There is a lot to unpack with support for a new model. Doing research and looking to document findings but ultimately believe the majority of complexity will be wrapped into processing of a profile for use with another model (either upstream in a catalog or downstream in a component/ssp). 

Intent is to document and further decompose both profile issues as well as oscal model issues. 

## Related Issue

Fixes #67 

Relates to #

## Type of change

## Checklist before merging


docs/cli-commands/lula_generate_profile.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+### Options
+
+```
+  -e, --exclude strings      comma delimited list of controls to exclude from the source catalog/profile
+  -h, --help                 help for profile
+  -i, --include strings      comma delimited list of controls to include from the source catalog/profile
+  -o, --output-file string   the path to the output file. If not specified, the output file will be directed to stdout
+  -s, --source string        the path to the source catalog/profile
```
meganwolf0, line 41: Is this the same thing as the -c/--catalog from generate component? Maybe using the same flags would be good there if so

src/cmd/generate/generate.go
```diff
@@ -134,6 +135,85 @@ var generateComponentCmd = &cobra.Command{
 	},
 }
 
+var profileExample = `
```
meganwolf0: Total organizational nit, should we split the profile generation cmd out to a separate file? I could see the generate starting to grow and might make it a bit easier from like a "finding things" perspective? Maybe like a `generate` root that adds the commands? Food for thought.

src/pkg/common/oscal/profile.go
```diff
@@ -0,0 +1,184 @@
...
+
+		// Does not handle pattern matching
+		for _, item := range importItems {
+
+			// Shouldn't be both but we can still handle the scenario
+			if item.IncludeControls != nil {
+				includeControls := *item.IncludeControls
+				for _, includeControl := range includeControls {
+					includes := *includeControl.WithIds
+					sort.Slice(includes, func(i, j int) bool {
```
meganwolf0: More of an FYI if we want to implement - I added a function that I'm using in the console to sort controls, that does like a smarter sort if they're XX-##.## format
```
slices.SortStableFunc(controls, func(a, b control) int {
	return oscal.CompareControlsInt(a.title, b.title)
})
```

src/cmd/generate/generate.go
```diff
@@ -134,6 +135,85 @@ var generateComponentCmd = &cobra.Command{
...
+
+			return nil
+		},
+	}
+
+	profilecmd.Flags().StringVarP(&source, "source", "s", "", "the path to the source catalog/profile")
+	profilecmd.MarkFlagRequired("source")
+	profilecmd.Flags().StringVarP(&outputFile, "output-file", "o", "", "the path to the output file. If not specified, the output file will be directed to stdout")
+	profilecmd.Flags().StringSliceVarP(&include, "include", "i", []string{}, "comma delimited list of controls to include from the source catalog/profile")
+	profilecmd.Flags().StringSliceVarP(&exclude, "exclude", "e", []string{}, "comma delimited list of controls to exclude from the source catalog/profile")
```
meganwolf0: I think there was a comment down in the profile section, that `include` and `exclude` shouldn't be specified at the same time - I think `profilecmd.MarkFlagsMutuallyExclusive("include", "exclude")` might work here?

src/test/e2e/cmd/generate_profile_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/defenseunicorns/lula/src/cmd/generate"
+	"github.com/defenseunicorns/lula/src/pkg/common/oscal"
+)
+
+func TestGenerateProfileCommand(t *testing.T) {
```
meganwolf0, line 14: I tried generating a profile with some invalid include/exclude controls, and it didn't scrub those to say they weren't in the source/catalog. IDK how the `include-control`/`exclude-controls` parameter works exactly (or if it would even falter if invalid control IDs were put there) but just wondering if we should include a check on if those control IDs are in the catalog or not.

meganwolf0 (changes requested):
probably going to play with it more/give more thought to the OSCALModel interface, but just to kick back some thoughts as a starter

docs/cli-commands/lula_generate_profile.md
brandtkeller: This is proposing a singular flag to be used for both profile/catalog sources and a previously captured request #563 would align generate component. 

Thoughts?

src/cmd/generate/generate.go
brandtkeller: I agree - will separate.

src/test/e2e/cmd/generate_profile_test.go
brandtkeller: I made a note about this in the doc I forgot to create initially. Right now we do no validation that controls specified exist in the source. We don't ingest the source at all currently. This just gets you from A- > B with profile creation. 

Besides the validation of existence - nothing is ultimately transient between the source and profile. This validation will be common as we port component generation over to using `source` as well.

src/cmd/generate/generate.go
brandtkeller: Added! great call out for cobra implementation improvements.

src/pkg/common/oscal/profile.go
brandtkeller: Added!

src/cmd/generate/profile.go
```diff
@@ -0,0 +1,91 @@
...
+				outputFile = "profile.yaml"
+			}
+
+			// pre-check if the output file exists
+			exists, err := pkgCommon.CheckFileExists(outputFile)
+			if err != nil {
+				return err
+			}
+
+			if exists {
```
CloudBeard: I don't think this is a prevent PR item but I noticed you can't re-run the generate command to overwrite a profile you created.

IE running the following command twice

`bin/lula generate profile -s https://raw.githubusercontent.com/GSA/fedramp-automation/refs/tags/fedramp-2.0.0-oscal-1.0.4/dist/content/rev5/baselines/yaml/FedRAMP_rev5_MODERATE-baseline-resolved-profile_catalog.yaml -i ac-1,ac-2,ac-3`

Or running either combo of the following commands

`bin/lula generate profile -s https://raw.githubusercontent.com/GSA/fedramp-automation/refs/tags/fedramp-2.0.0-oscal-1.0.4/dist/content/rev5/baselines/yaml/FedRAMP_rev5_MODERATE-baseline-resolved-profile_catalog.yaml -i ac-1,ac-2,ac-3`

`bin/lula generate profile -s https://raw.githubusercontent.com/GSA/fedramp-automation/refs/tags/fedramp-2.0.0-oscal-1.0.4/dist/content/rev5/baselines/yaml/FedRAMP_rev5_MODERATE-baseline-resolved-profile_catalog.yaml -i ac-1,ac-2`

You'll get `Error: output File profile.yaml currently exist - cannot merge artifacts`

You can do that with component-definitions is the only reason I noticed.

CloudBeard:
I know this isn't OSCAL Schema compliant but Lula functionality wise what do you think about a situation where you want to create a profile for IL4 using FedRAMP Moderate that would introduce the need to add the control AU-5(1) which isn't in the baseline so that control should come from the FedRAMP High Baseline.

Im thinking command/functionality wise allowing multiple sources, have something like `-ia` (include-all, for Moderate baseline), and `-i` for the specific controls in High Baseline.

Reconcile what that would actually look like in OSCAL but a quick way to say give me all of FedRAMP Moderate and this control(s) from High. 


(Also don't think this blocks this PR, its more of a future request/thought)

docs/cli-commands/lula_generate_profile.md
meganwolf0: Ok after looking at the issue and back at the commands... is the logic to handle either (to populate the component or profile) different? Would it be better from a consuming that source perspective to have those indicated differently (just so you avoid having to try and read the doc not knowing up front what it is)? Given my OSCAL-ese is pretty bad, I'd defer to you all on that one. 

Doesn't look like component currently supports Profile, although the flag is available, so I can't really tell the difference between what happens under the hood if either is chosen...

meganwolf0: And if I scroll down to the next comment, it sounds like maybe that's the answer to above - the source doesn't really matter?

brandtkeller: The idea with source as used in profile and component definition is that it can point to either a profile or a catalog. 

So we have a TODO which is to support component-definitions pointing to either (and not knowing which until we read it into the complete schema. If we write this logic to support component-definitions - it can be re-used here in the future as the same logic. 

For the current scope - we're not doing any processing of what source is as of yet. In a follow on issue (i'll be decomposing these shortly) we'll be performing a profile-resolve which will need to utilize source. I wanted to keep this separate as it will introduce a lot of new logic.

src/cmd/generate/profile.go
brandtkeller: Correct! `HandleExisting` is going to currently return an error for the profile model. There is a lot of complexity baked into what you can do with a profile and I believe initially I found merging to be a constraint. 

Thinking on it more and there will certainly be things that don't make sense with generation but merging might not be terrible. I'll give that a shot.

brandtkeller: Added support for merging behaviors with 1 caveat:

Instead of trying to support every single permutation possible of include/exclude/all -> when a conflicting `import` is found during merge - the latest will take precedence and overwrite the original import of the same `href`. 

Maybe we can support more granularity in the future - but I'd like to see if/how this command is used.

brandtkeller:
> I know this isn't OSCAL Schema compliant but Lula functionality wise what do you think about a situation where you want to create a profile for IL4 using FedRAMP Moderate that would introduce the need to add the control AU-5(1) which isn't in the baseline so that control should come from the FedRAMP High Baseline.
> 
> Im thinking command/functionality wise allowing multiple sources, have something like `-ia` (include-all, for Moderate baseline), and `-i` for the specific controls in High Baseline.
> 
> Reconcile what that would actually look like in OSCAL but a quick way to say give me all of FedRAMP Moderate and this control(s) from High.
> 
> (Also don't think this blocks this PR, its more of a future request/thought)

I was going to move it off to another issue but you gave me to motivation to support an `--all` flag for importing all controls. 

I believe we can accomplish the scenario above with a few commands. Processing that will be another story and scope for another issue I will write.

meganwolf0:
> I was going to move it off to another issue but you gave me to motivation to support an `--all` flag for importing all controls.

Trying to follow the train of thought in the thread, but would --all not be the default if include or exclude wasn't provided? I guess, what happens in no flag vs --all?

brandtkeller:
> Trying to follow the train of thought in the thread, but would --all not be the default if include or exclude wasn't provided? I guess, what happens in no flag vs --all?

Great point - that was a gap. Added a OneRequired for the flags to lean into explicit vs default.

meganwolf0 (dismissed):
Approved! Are we going to create follow-on issues to reconcile the `OSCALModel` interface across the other models we have?

meganwolf0 (approved):

brandtkeller:
> Approved! Are we going to create follow-on issues to reconcile the `OSCALModel` interface across the other models we have?

Added issues as referenced a few hours ago

PR #696, opened 2024-10-01
chore: add global command context for program cancelation (and anything else)

### LINKED ISSUE #682: Support for cancellation
While executing lula commands, when the user hits ctrl-c or running processes exceeds a (configurable) timeout, we should be able to cancel the operation and clean up any resources changed by executable domains.  

If we wanted (or are pressed for time) I think starting with a basic context with cancel at the CLI level that gets passed to domains & providers would be sufficient, the timeouts and configuration can come later (feel free to disagree).

mildwonkey:
## Description

This implements cancelation with CTRL-C and adds the context with cancelation to the cobra root command. The context is not used anywhere else yet, but sub-commands can access it in their Run methods w/ `cmd.Context()`.

closes #682 

## Type of change

## Checklist before merging

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -58,16 +58,10 @@ github.com/charmbracelet/bubbletea v1.1.1 h1:KJ2/DnmpfqFtDNVTvYZ6zpPFL9iRCRr0qqK
 github.com/charmbracelet/bubbletea v1.1.1/go.mod h1:9Ogk0HrdbHolIKHdjfFpyXJmiCzGwy+FesYkZr7hYU4=
 github.com/charmbracelet/lipgloss v0.13.0 h1:4X3PPeoWEDCMvzDvGmTajSyYPcZM4+y8sCA/SsA3cjw=
 github.com/charmbracelet/lipgloss v0.13.0/go.mod h1:nw4zy0SBX/F/eAO1cWdcvy6qnkDUxr8Lw7dvFrAIbbY=
-github.com/charmbracelet/x/ansi v0.2.3 h1:VfFN0NUpcjBRd4DnKfRaIRo53KRgey/nhOoEqosGDEY=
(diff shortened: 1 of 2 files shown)


meganwolf0 (commented):
I mostly just have some questions - I tested this branch compared to behavior of main with ctrl+c and I'm not exactly sure I'm seeing the difference in behavior, other than printing the output message (but also only tested a couple things so might be missing it). 
- I was slightly concerned about this overriding some of the ctrl+c handling performed in the tui, but that seems to not affect things, so that's good
- Is the intent of this to like more gracefully handle random cancel/shutdown operations? I tested this with a validate for pepr (which takes a while because it's creating resources) and the ctrl+c just kills everything with no existing resource cleanup, which might be a separate thing to address -> I guess idk if the purpose of this function is to leave space for something like that to be instrumented

(Really just looking for an ELI5 here 😄 )

mildwonkey:
> Is the intent of this to like more gracefully handle random cancel/shutdown operations?

Yes, exactly - by putting this at the root command, we *can* add graceful cancellation to all other commands (as needed/desired). I purposefully didn't change anything else's behavior, just added the functionality, so at the moment (as you already noticed) this just cuts commands short; we still need to implement the graceful shutdown behaviors. 

I also have a secondary reason for adding the context that will come in a future PR: I'd like to explore setting context keys and use the command context to pass info:
```ctx := context.WithValue(cmd.Context(), types.LulaValidationWorkDir, filepath.Dir(opts.InputFile))```

But (imo) using a context to support graceful shutdown throughout all of lula's commands is reason enough for this, even if we don't go the context key route.

brandtkeller (approved):
While this PR and the Issue have a (proper) scope - I've wanted to explore many ways with which to use context for both scenarios such as this and other patterns I've seen used in CLI apps.

meganwolf0 (approved):

PR #697, opened 2024-10-02
chore: move SetCwdToFileDir to new `workdir` package

mildwonkey:
so it can be imported into the file domain (otherwise there were import cycle issues when importing "common"). I'm entirely open to different package names, feel free to nitpick that!

I split this out into it's own PR because the file domain PR is a big mess and I know that there will be some merge conflicts with other in-flight PRs. I'm happy to wait for other PRs to merge and fix conflicts in this one.


meganwolf0:
Might be beneficial to sync up on this one - in the composition PR I totally removed the cwd changes, if that needs to be restored and solved another way because it really throws a wrench in the file domain work stuff though let me know - I think we had some slack discussions where we wanted to remove that cwd jumps, but perhaps need some justification if it's too hard to make other things close...

mildwonkey:
I can handle resolving file locations differently, I was just trying to use what we had in place already. 

Resolving filepaths is _easiest_ if I can handle it by cd'ing into the directory that the paths are relative to, but it's not strictly necessary and if we're getting away from the directory-changing behavior altogether I'll handle it differently. (I don't like having to calculate absolute filepaths because once you get into different OSes and add in symlinks some things get really buggy really quickly, but it's all doable)

PR #698, opened 2024-10-02
fix: refactor, retries, sleep to address flaky tests

### LINKED ISSUE #689: Flaky CI Unit Tests for Console
### Description

The unit testing for the TUI/console functionality (using the bubbletea teatest library) is constantly flaky in CI - generally having to re-run the job several times before it passes. The `fm` (FinalModel output) is returning nil. 

Bumping timeout has not really helped, need to uncover other library configurations to toggle and/or if we want to conditionally run these tests in CI (which the model changes)

### Additional Context

Determine short/long term plan to address

meganwolf0:
## Description

Did some refactor to try and call submodels, as opposed to the entire model - this doesn't make too much of a difference at the moment, but as the model support grows + as we import more oscal, running all the routines could be costly.

Major change was creating a test model view helper, then implementing retries + sleeps between msg send commands, I noticed that with the picker model intro one of the tests regressed and it slipped through (basically it was happening too quickly that the picker didn't have time to load and the components weren't being swapped in the view)

## Related Issue

Relates to #689 

## Type of change

## Checklist before merging


meganwolf0:
for what it's worth, this passed unit tests the first go-around (could be a coincidence though)

brandtkeller (approved):
Helper function adds resilience to the testing process and standardizes workflows. Organization of the test data is intuitive.

PR #703, opened 2024-10-03
feat(domains): file domain

### LINKED ISSUE #337: Support for a file domain

mildwonkey:
## Description
This PR adds a `files` domain which use's OPA's conftest parser to read in a long list of configuration file types* so that rego validations can be written against the contents. This domain works with OPA and the kyverno provider, though for kyverno (as far as I understand) you need to use the kyverno-json module. 

Domain example:
```yaml
domain:
  type: file
  file-spec:
    filepaths:
    - name: grafana-config
       path: custom.ini
```

The file domain takes an array of files, but the implications are that the output DomainResources map uses the file name as a key, so you need to include the filename in the validation:

```yaml
      package validate
      import rego.v1

      # Default values
      default validate := false
      default msg := "Not evaluated"
      
      validate if {
       check_grafana_config.result
      }
      msg = check_grafana_config.msg
      config := input[grafana-config] # <-- MAP KEY HERE!
      protocol := config.server.protocol

      check_grafana_config = {"result": true, "msg": msg} if {
        protocol == "https"
        msg := "Server protocol is set to https"
      } else = {"result": false, "msg": msg} if {
        protocol == "http"
        msg := "Protocol is http. That's bad, mmkay?"
      }
```

Original PR for reference: https://github.com/defenseunicorns/lula/pull/688

Relates to #337

Note to reviewers: besides all the obvious new files, the major _change_ in here is the introduction of a context.Context that starts from the validate command and flows through to `GetResources()` - the actual place that the file domain needed the information in question. I've added a context key named `LulaValidationWorkDir` (NITPICK FREELY! NAMING THINGS IS HARD!) and I'm using that to store the location that the lula command is run from, so that the file domain can build a relative path (if necessary; it also supports abs paths and network downloads) to location the files to evaluate. I kept it as simple as I could and only added the context where necessary, but I think there's potential to use the command context (and additional context keys) as lula command request contexts throughout the codebase. 

----------------------------
From https://www.conftest.dev/:

```
As of today Conftest supports:

CUE
CycloneDX
Dockerfile
EDN
Environment files (.env)
HCL and HCL2
HOCON
Ignore files (.gitignore, .dockerignore)
INI
JSON
Jsonnet
Property files (.properties)
SPDX
TextProto (Protocol Buffers)
TOML
VCL
XML
YAML
```
--------------------------
## Type of change

## Checklist before merging

- [X [Contributor Guide Steps](https://github.com/defenseunicorns/lula/blob/main/CONTRIBUTING.md) followed


docs/reference/domains/file-domain.md
```diff
@@ -0,0 +1,125 @@
...
+
+      config := input["grafana"]
+      protocol := config.server.protocol
+
+      check_grafana_config = {"result": true, "msg": msg} if {
+        protocol == "https"
+        msg := "Server protocol is set to https"
+      } else = {"result": false, "msg": msg} if {
+        protocol == "http"
+        msg := "Grafana protcol is insecure"
```
brandtkeller:
```suggestion
        msg := "Grafana protocol is insecure"
```

brandtkeller (changes requested):
besides a typo my only request would be an unhappy-path E2E test where the data residing in the associated file does not adhere to the policy. I believe you hinted to this with the https vs http example. I don't believe I saw an unhappy path test.

docs/reference/domains/file-domain.md
mildwonkey: Thanks for the test pointer - that also let me realize I left out the kyverno test (the fixture made it in 🙄, not the test)~

mildwonkey:
Added test coverage for kyverno (whoops!), not-satisfied validations, and one for invalid inputs (just an empty file; just proof we don't panic)

brandtkeller (dismissed):
No remaining items worth blocking this from merge. 

@meganwolf0 - would appreciate a second pair of eyes. 

Otherwise two items of opportunity I'll call out:
1. We've been discussing the idea of `E2E` tests becoming more CLI-driven in the near future. I believe some of that structure is still under review but I wouldn't be surprised if we swap these E2E tests in the future.
2. Deciding on the role of entangling templating into our tests in order to reduce the file surface. IE for both pass/fail the same component definitions are used. If we templated the path of the config file - we could probably de-duplicate. Again - not a feature that is ready for use just yet. 

Food for thought, otherwise appreciate this addition!

docs/reference/domains/file-domain.md
```diff
@@ -0,0 +1,125 @@
+# File Domain
+The File domain allows for validation of arbitrary file contents. The file domain can evaluate local files and network files. Files are copied to a temporary directory for evaluation and deleted afterwards.
+
+## Specification
+The File domain specification accepts a descriptive name for the file as well as it's path. The names must be unique.
```
meganwolf0:
```suggestion
The File domain specification accepts a descriptive name for the file as well as its path. The names must be unique.
```

docs/reference/domains/file-domain.md
```diff
@@ -0,0 +1,125 @@
...
+domain:
+  type: file
+  file-spec:
+    filepaths:
+    - name: config
+      path: grafana.ini
+```
+
+## Supported File Types
+The file domain use's OPA's [conftest](https://conftest.dev) to parse files into a json-compatible format for validations. ∑Both OPA and kyverno (using [kyverno-json](https://kyverno.github.io/kyverno-json/latest/)) can validate files parsed by the file domain.
```
meganwolf0:
```suggestion
The file domain uses OPA's [conftest](https://conftest.dev) to parse files into a json-compatible format for validations. Both OPA and kyverno (using [kyverno-json](https://kyverno.github.io/kyverno-json/latest/)) can validate files parsed by the file domain.
```

brandtkeller (dismissed):
Deconflicted some recent updates to go mod/sum

meganwolf0:
Not sure if I'm not setting things up right, but still not getting the compose behavior I'd expect with the root directory/modified path in the filepath (attached my test files)
[files_domain.zip](https://github.com/user-attachments/files/17281726/files_domain.zip)

The `comp-def-composed` is the result of `lula t compose -f comp-def.yaml` - I was expecting the filepath.path to be relative to the `comp-def-composed` location. 

Also when I run `validate` I'm still getting the GetResources error: `Error running validation: domain GetResources error: error getting source files: open /Users/meganwolf/Documents/defense-unicorns/lula-compliance-lib/grafana.ini: no such file or directory` so I think this probably needs to be addressed in the underlying composition functionality

meganwolf0 (changes requested):

meganwolf0:
(Ok, I added this comment, then deleted when I thought I tested the wrong branch 😅 but I think I got the right one so still stands)
Here are my test files, happy to hear I'm screwing something up in the set up! 
[files_domain.zip](https://github.com/user-attachments/files/17282215/files_domain.zip)

Basically, when I compose and validate, I'd expect the `filepaths[_].path` to be relative with respect to where the component definition is - since we might have the scenario where we'd have nested component definitions (e.g., `import-component-definitions`) or just remote validations.

mildwonkey:
@meganwolf0  - thanks for the example files. I wasn't accounting for compose at all (specifically the multiple levels of nested file path).  I'll dig into refactoring compose to support the file domain and make sure this is covered in the e2e tests.

meganwolf0:
One more thing, and I think this is really more of a team convo (@brandtkeller) but we've called all the files that contain our `types` in the domains/providers `types.go` (personally I like `spec.go`, feels more descriptive on what's contained therein) but I suppose for consistency sake, should we use `types`? Or should we say going forward we'll use spec and probably create an associated re-naming issue of some of the files in the existing domains/providers to be more descriptive/consistent?

mildwonkey:
That's a good point, I did notice that y'all use naming conventions I'm not used to. I don't object to renaming files (even to "types"), but i do want to say that lula's use of "types" is very unintuitive to me. I found files with actual code implementation in a file just called "types", where I rather assumed I'd find at most an interface definition.

mildwonkey:
> Deconflicted some recent updates to go mod/sum

😓 whoops @brandtkeller thank you but fwiw if you push changes to my PRs I will probably not see the changes before force pushing over them after a rebase

mildwonkey:
I've updated the docs to match our recent conversation: the file domain works with relative file paths but if you're using compose, filepaths have to be relative to the output directory/working directory. (I don't like how I worded it here, and I don't like how I worded it there, so please nitpick!)\

@brandtkeller - @meganwolf0 had a good question in her last comment, about naming the "extra" file in the files domain `types.go` instead of `spec.go`. I'm absolutely willing to rename it if y'all want, though I am personally strongly opposed to the naming convention. thoughts?

brandtkeller:
> @brandtkeller - @meganwolf0 had a good question in her last comment, about naming the "extra" file in the files domain `types.go` instead of `spec.go`. I'm absolutely willing to rename it if y'all want, though I am personally strongly opposed to the naming convention. thoughts?

If you believe this is better convention and current guidance only points at consistency as an argument then I'm onboard with keeping proposed convention as long as we get an issue or some traceable artifact that points to eventually aligning said conventions.

docs/reference/domains/file-domain.md
```diff
@@ -0,0 +1,128 @@
...
+* Property files (.properties)
+* SPDX
+* TextProto (Protocol Buffers)
+* TOML
+* VCL
+* XML
+* YAML
+
+## Validations
+When writing validations against files, the filepath `Nam`e must be included as
```
brandtkeller:
```suggestion
When writing validations against files, the filepath `Name` must be included as
```

brandtkeller (commented):
Minor typo otherwise new documentation supports discussion and continuation once defined in #715 and #716

meganwolf0 (approved):

brandtkeller (approved):

PR #704, opened 2024-10-04
fix(validate): support returning empty resources object/slice

### LINKED ISSUE #589: Allow domain to return zero resources without failing

In thinking about general workflows for best-practices and STIG's, Lula may want to support the ability to return zero resources from a domain data collection activity. 

Currently the inability to find > 0 resources will fail a Validation as `not evaluated` as there were no items found. 

I thinking about something such as a best practice or STIG - there are at least two primary scenarios:
1. If a resource exists -> perform policy evaluation to determine adherence
2. If a resource does not exist -> there is no applicable evaluation to be done (IE success)

both of these should be drive from the provider policy instead of the domain. This allows the validation author to write a policy that either passes or fails when zero resources are retrieved.

brandtkeller:
## BLUF

Domains are responsible for the collection of data that will be processed by a provider. This PR makes a stance that domains should:
- Always perform the full collection of data possible(do not exit on error before completion)
- Return an empty object when an error occurs

This sets a standard where Lula always attempts to retrieve the full extent of evidence possible. When an error occurs - reviewers/auditors should want to have a full view of what worked and what did not with regards to domain resource collection. 

## Description

Looking to setup the changes required to test impact to our current validations on allowing domains to return a map item that contains an empty payload. Most applicable to our k8s domain but still relevant elsewhere. 

> Update: We will need to rewrite some of our existing policies to account for an edge case where no resources were collected. This is more of an edge case in my opinion as the stability of the current validations has been good.

Context primarily captured in the issue below - but the primary goal is that providers should be gating whether the data allows for policy adherence and not a domain. A domain should be objective in that - if there was no errors during the collection of the data - it should continue to progress. 

Removed a return statement in `QueryCluster` and instead aggregate the errors into a single string - Given the ability to continue writing resources if one of the `resource` items fails - this would allow for the validation to still collect evidence for later review - I believe this is a fundamental part of how Lula should operate - always being in a position to provide as much evidence as possible to enable less guessing in the event a validation or environment needs correction.

## Related Issue

Fixes #589

## Type of change

## Checklist before merging


src/pkg/domains/kubernetes/resource.go
```diff
@@ -23,24 +24,39 @@ func QueryCluster(ctx context.Context, cluster *Cluster, resources []Resource) (
 	// We may need a new type here to hold groups of resources
 
 	collections := make(map[string]interface{}, 0)
+	var multierr []error
```
mildwonkey: note: This isn't a merge-blocking question
I'm a fan of multierror, but I've switched to using `errors.Join()` since the wrap/unwrap behavior is close enough to the same and it's stdlib ([example in api PR](https://github.com/defenseunicorns/lula/pull/766/files#diff-6097654b3efc7b30cd833368b6928abd5aa00c4e7afc4f656cc1ed3976b32bbbR16). What do you think?

src/pkg/domains/kubernetes/resource.go
```diff
@@ -23,24 +24,39 @@ func QueryCluster(ctx context.Context, cluster *Cluster, resources []Resource) (
...
+				// This request returned no resources
+				collections[resource.Name] = []map[string]interface{}{}
 			}
 		}
 	}
+
+	if multierr != nil {
+		var result *multierror.Error
+		result = multierror.Append(result, multierr...)
+		return collections, fmt.Errorf("%s", result.Error())
```
mildwonkey: You should be able to return the result here without the fmt statement, it's already satisfying the error interface

mildwonkey (dismissed):
I left some comments, but none of them have to block merging this as-is from my POV.

src/pkg/domains/kubernetes/resource.go
brandtkeller: Love it! Glad to see you noticed this same pattern and addressed in the in API work - I'll switch over.

src/pkg/domains/files/files.go
```diff
@@ -68,23 +76,32 @@ func (d Domain) GetResources(ctx context.Context) (types.DomainResources, error)
 	// get a list of all the files we just downloaded in the temporary directory
 	files, err := listFiles(dst)
 	if err != nil {
```
brandtkeller, line 78: if `listFiles()` returns an error should we return what we have?

src/pkg/domains/files/files.go
```diff
@@ -112,7 +130,8 @@ func (d Domain) GetResources(ctx context.Context) (types.DomainResources, error)
 		// get a list of all the files we just downloaded in the temporary directory
 		files, err := listFiles(parserDir)
 		if err != nil {
-			return nil, fmt.Errorf("error walking downloaded file tree: %w", err)
+			errs = errors.Join(errs, fmt.Errorf("error walking downloaded file tree: %w", err))
+			return drs, errs
```
mildwonkey: nitpick (but relevant for this file, i think): since you are returning here there's no need to join errs on their own line

`return drs, errors.Join(errs, fmt.Errorf...) `

It's nitpicky and you don't need to make a change, but it is kind of misleading: given the rest of this file, my first assumption on seeing you return right after joining to a chain of errors is that it's a bug and you didn't mean to return there.

src/pkg/domains/files/files.go
mildwonkey: I would return what we have - it's an accurate zero value.

caveat: I would also expect that an error here would expect would stop lula from running validate, but Lula currently swallows errors returned from GetResources if the domain was empty - I'm changing that behavior in a PR, but if you merge this right now you might see weird behavior .

proof/change: https://github.com/defenseunicorns/lula/pull/766/files#diff-2b1d27b989848257f397062966fca3725bac1057960904196ddcf28d17b0b73dR155

brandtkeller: Appreciate the sanity check. 

Just to double back on your comment - we wouldn't exit a `validate` execution entirely due to one malformed `validation` yes?

Each validation should be a wholistic proof of evidence not coupled to other `validations` - Our stance on this is that there is still reason to complete the analysis of other controls/validations such that you do not require a full re-run in order to evaluate all areas of further evaluation.

src/pkg/domains/files/files.go
brandtkeller: Proper nitpick IMO

src/pkg/domains/files/files.go
mildwonkey: Agreed, I think the goal is to return as many actual results as possible (I'd love to have a side conversation about "not-satisfied" results where something like "ERRORED" would be more informative, but that's neither here nor there)

mildwonkey (dismissed):
I have some feedback but I don't think it's anything that has to block merging this as-is. I'm a bit concerned that changing this behavior didn't necessitate test changes; we should consider adding some tests that validate the partial response on error.

meganwolf0:
I think this looks good, possibly agree with Kristin's comment above. Also wondering if you did any current validation tests and measured the outcomes against this branch v main (just to make sure partial resource return isn't flipping satisfaction on any existing validations, particularly those in UDS Core)? 

I feel like this is a good test case for figuring out how to really instrument a proper testbed for that sort of thing - obviously we cant test every validation, but might be nice to test a large swath - or even have an action that we support to test old v new Lula version that we could provide people to run in their own environments to ensure that their validations are still responding as expected with Lula version upgrades...

brandtkeller:
> I think this looks good, possibly agree with Kristin's comment above. Also wondering if you did any current validation tests and measured the outcomes against this branch v main (just to make sure partial resource return isn't flipping satisfaction on any existing validations, particularly those in UDS Core)?

Tested against uds-core and did not reduce counts of `satisfied` controls. 

> I feel like this is a good test case for figuring out how to really instrument a proper testbed for that sort of thing - obviously we cant test every validation, but might be nice to test a large swath - or even have an action that we support to test old v new Lula version that we could provide people to run in their own environments to ensure that their validations are still responding as expected with Lula version upgrades...

I like that idea!

meganwolf0 (approved):

PR #708, opened 2024-10-04
feat(validate): template oscal during runtime

### LINKED ISSUE #592: Template OSCAL during runtime

Ability to template OSCAL files during execution (validate)

- **Given** an OSCAL file with variable declarations exists
- **When** a validate is executed
- **Then** Lula templates variables into the OSCAL

What is really required for templating? generic capability may provide future value. 

### Additional context

Composing may end up being the real template value for OSCAL. But this being the more complex route could benefit from the exploration

### Expected Deliverable

Ability to inject some modification of OSCAL with a templated variable during `validate`

meganwolf0:
## Description

Supporting templating during `lula validate`

* Added a ValidationContext (similar to CompositionContext) to hold the validation-specific parameters for performing a validation
* Refactor of the `cmd/validate` functions -> `src/pkg/validation`
* No `lula validate` cli changes, just assumed anything that can be templated will be done so by default

Optional follow-on to this would be to save off the `composed` file that is `validated` - we discussed this previously but I did not include here to simplify.

## Related Issue

Fixes #592 

## Type of change

## Checklist before merging


src/test/e2e/api_validation_test.go
```diff
@@ -55,7 +55,12 @@ func TestApiValidation(t *testing.T) {
 			oscalPath := "./scenarios/api-field/oscal-component.yaml"
 			message.NoProgress = true
 
-			assessment, err := validate.ValidateOnPath(context.Background(), oscalPath, "")
+			validationCtx, err := validation.New()
+			if err != nil {
+				t.Errorf("error creating validation context: %v", err)
+			}
+
+			assessment, err := validationCtx.ValidateOnPath(context.Background(), oscalPath, "")
```
mildwonkey: 🤔  this naming convention is still really confusing; I looked at this and initially didn't understand why you weren't writing `ValidateOnPath(validationCtx, oscalPath, "")`  

This is one of the two main use cases for the `context.Context()` that I added to the root commands. It's good for passing request-specific contextual values. Is there a reason you didn't use that? 

If you don't want to use the `ctx` (which is valid, I just want to understand why we wouldn't use what's already there for that purpose), can we rename these so it's not quite so ambiguous? "context"  has a very specific connotation in go and naming a struct "anythingctx" or "anythingcontext" leads me to assume it is a `context.Context`, so I prefer not to use that in other ............... contexts. 😭

meganwolf0: Yeah we can call it something else instead of context, I was hoping you'd word-smith me something haha but we didn't really ever close it on compose so just left it as is... but idk, like ComposeBuilder/ValidationBuilder? I'm not a creative type.

> This is one of the two main use cases for the context.Context() that I added to the root commands.

Which part would I substitute out for using the context from the root? Not totally following there

src/pkg/common/composition/resource-store.go
```diff
@@ -157,3 +159,18 @@ func (s *ResourceStore) fetchFromRemoteLink(link *oscalTypes_1_1_2.Link, baseDir
 
 	return ids, err
 }
+
+func createTemplateResource(data []byte) *oscalTypes_1_1_2.Resource {
```
brandtkeller: this is just a historical artifact that needs cleanup correct?

brandtkeller (changes requested):
I don't see much that looks unsuited to the task. Deconflicting with main and some cleanup - otherwise looks good.

src/pkg/common/composition/resource-store.go
meganwolf0: ah it snuck back in, removed!

src/test/e2e/api_validation_test.go
meganwolf0: renamed!

brandtkeller (dismissed):

brandtkeller (approved):

PR #714, opened 2024-10-08
feat(kubernetes): support running both create resources and resources in the kubernetes spec

### LINKED ISSUE #531: Support running both create-resources and resources in the kubernetes-spec
Right now, you can only execute `create-resources` or `resources` when getting resources for a Lula Validation. It would be helpful to adjust that routine so that we could `create-resources`, then get different resources. Currently `create-resources` only returns those that are actually created.

### Additional context
The use case arose such that we might want to deploy a sample application and measure creation of network policies by the UDS Controller, for instance.

meganwolf0:
## Description

Adds operations to support create -> wait -> read
* Fixed a bug that I found with the ExecuteWait (apparently the object returned from `validateAndGetGVR` doesn't have the correct Group value, so more learning possibly there as to why - so this had passed with a basic "pod" but not with something more complex like a deployment)
* Re-arranged KubernetesDomain `GetResources` to evaluate CreateResources -> Wait -> Resources, combining the create and resources into a final `types.DomainResources`. Based on the combining (just a simple map merge) the keys (names) between the two need to be unique
* Finally, renamed `types` -> `spec` following the discussion on naming clarity

## Related Issue

Fixes #531 

## Type of change

## Checklist before merging


brandtkeller (approved):
Appreciate identifying a bug and addressing it with a new test.

PR #718, opened 2024-10-09
feat(kubernetes)!: wait logic kubernetes version support

### LINKED ISSUE #229: Wait logic kubernetes version support
The wait logic will break on (at a minimum) Kubernetes version 1.24.

Review the current wait workflow and determine version support OR remove logic that requires certain kubernetes versions.

meganwolf0:
## Description

* Modifies the connect and wait logic to use [unicorn k8s pkg](https://github.com/defenseunicorns/pkg/blob/main/kubernetes/wait.go) - should address the k8s version support bug that the core of this issue is concerned with
* Refactors creation of `cluster` so as to not duplicate `connect` invocations
* Proposal of modification to `wait` schema to be more aligned to `resource-rule` (since all the `kind` logic is no longer needed - also makes it easier to validate user inputs before passing to wait routine):

```yaml
domain:
  type: kubernetes
  kubernetes-spec:
    wait:                               # Optional - Group of resources to read from Kubernetes
      group:                            # Optional - Empty or "" for core group
      version: v1                       # Required - Version of resource
      resource: pods                    # Required - Resource type (API-recognized type, not Kind)Required - Resource type (API-recognized type, not Kind)
      name: test-pod-wait               # Required - Name of the resource to wait for
      namespace: validation-test        # Optional - For namespaced resources
      timeout: 30s                      # Optional - Defaults to 30s
    resources:
    - name: podsvt
      resource-rule:
        group:
        version: v1
        resource: pods
        namespaces: [validation-test]
```

## Related Issue

Fixes #229

## Type of change

- [X] Bug fix (breaking change which fixes an issue)

## Checklist before merging


src/pkg/domains/kubernetes/cluster.go
```diff
@@ -0,0 +1,95 @@
...
+	kclient       klient.Client
+	watcher       watcher.StatusWatcher
+	dynamicClient *dynamic.DynamicClient
+}
+
+func GetCluster() (*Cluster, error) {
+	clusterConnectOnce.Do(func() {
+		globalCluster, globalConnectionErr = New()
+	})
+	if globalConnectionErr != nil {
```
mildwonkey: I think the error checking should happen within the `Do`,  you shouldn't need to check that error every time.

brandtkeller (changes requested):
Appreciative of native existence support. Can we include updates the go mod/sum as I imagine there are items we can remove from our dependency surface with these changes.

src/pkg/domains/kubernetes/cluster.go
meganwolf0: good point, actually I think maybe I just probably don't need that error check at all? Now that I moved it inside the Do() I'm having a hard time seeing how it's useful, since Do doesn't return anything

brandtkeller (approved):

PR #726, opened 2024-10-10
feat(file domain): add support for reading arbitrary files as strings

### LINKED ISSUE #709: Files domain support for arbitrary file types

In efforts of enabling domains flexibility in the ability to rapidly prototype - we can enable users in the files domain who want to validate an unsupported file type. 

This request would allow un-supported file types to still be provided in a baseline payload to the provider in use. 

- **Given** arbitrary file types not included in the conftest library
- **When** A validation occurs
- **Then** the files domain provides the contents of the file to the provider in a consistent manner

A new filetype parser is required for each additional file type

### Additional context

Possibly just a string payload that can be operated on with Regex and/or other tooling/native functions.

### Expected Deliverable

- [ ] Ability to support arbitrary file types without any specific parsing

mildwonkey:
## Description
This PR adds support for arbitrary file contents read in as strings. Line endings are relevant when writing regexes, so I did absolutely nothing fancy when reading the file: go will preserver tabs and newlines, and users will need to be aware of that when writing validations. 

I will add more to the documentation about how to write a validation for a string-blob, but I wanted to see if the team had objections to the format before I do so. It's valid to say we're going to read it as a giant single string with no extra lines, tabs, or spaces, but that seemed unhelpful when trying to write a validation.

This also supports passing the name of the parser for any given file, not just the string file. The initial list is all of the conftest parsers supported as of the time of writing, plus our custom string parser. If the user specifies a `parser` in the file spec, that file will get special handling. Everything except the "string" parser gets passed into conftest's `ParseConfigurationAs` method with the parser name (I added it as an enum in the schema, so I'm not doing any other validation, but if I should add that for other callers I shall!), while strings get slurped into memory then added to the parsed configs map.  

I've added some tests (e2e and unit) for: 
- string files
- a json file without a file extension, using the json parser

I'm always willing to add tests, let me know what you'd like to see! I'll add some more words to the documentation once folks get a chance to comment on the string file format.  

## Related Issue

Closes #709 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist before merging

- [x ] [Schema Updates](https://github.com/defenseunicorns/lula/blob/main/docs/community-and-contribution/schema-updates.md) applied
- [ x] [Contributor Guide Steps](https://github.com/defenseunicorns/lula/blob/main/CONTRIBUTING.md) followed

### DIFF
--- a/src/test/e2e/scenarios/file-validations/pass/component-definition-string-file.yaml
+++ b/src/test/e2e/scenarios/file-validations/pass/component-definition-string-file.yaml
@@ -1,0 +1,89 @@
+component-definition:
+  uuid: E6A291A4-2BC8-43A0-B4B2-FD67CAAE1F8F
+  metadata:
+    title: OSCAL Demo Tool
+    last-modified: "2022-09-13T12:00:00Z"
+    version: "20220913"
+    oscal-version: 1.1.1
+    parties:
+      # Should be consistent across all of the packages, but where is ground truth?
+      - uuid: C18F4A9F-A402-415B-8D13-B51739D689FF
+        type: organization
+        name: Defense Unicorns
+        links:
+          - href: https://github.com/defenseunicorns/lula
+            rel: website
+  components:
+    - uuid: A9D5204C-7E5B-4C43-BD49-34DF759B9F04
+      type: software
+      title: lula
+      description: |
+        Defense Unicorns lula
+      purpose: Validate compliance controls
+      responsible-roles:
+        - role-id: provider
+          party-uuids:
+            - C18F4A9F-A402-415B-8D13-B51739D689FF # matches parties entry for Defense Unicorns
+      control-implementations:
+        - uuid: A584FEDC-8CEA-4B0C-9F07-85C2C4AE751A
+          source: https://github.com/defenseunicorns/lula
+          description: Validate generic security requirements
+          implemented-requirements:
+            - uuid: 42C2FFDC-5F05-44DF-A67F-EEC8660AEFFD
+              control-id: ID-1
+              description: >-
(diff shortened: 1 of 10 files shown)


src/test/e2e/scenarios/file-validations/pass/component-definition-string-file.yaml
```diff
@@ -0,0 +1,89 @@
...
+                validate if {
+                  check_server_protocol.result
+                }
+                msg = check_server_protocol.msg
+                
+                config := input["config"]
+                
+                check_server_protocol = {"result": true, "msg": msg} if {
+                  regex.match(
+                    `server = https\n`,
```
mildwonkey, line 74: ^^ this is the main bit I want to bring reviewer's attention to, the line endings. Given a file with multiple lines, our options are basically the following:

```
{"config": "server = https\nsomething = else"}
```

or:

```
{"config": "server = httpssomething = else"}
```

or:
```
{"config": "server = https something = else"}
```

(or we add weird new line breaks: 🦄 😉 )

I assumed we want the newlines - I, wearing my infra person hat, would expect them - but we could make it configurable or go with the no spaces / single space in place of new lines / option D / etc.

brandtkeller: I'd lean towards your intuition as well. Line endings may provide utility and my vote is generally for the least amount of intervention unless we receive demand signals for configuration or a new default stance.

meganwolf0 (dismissed):
all looks good to me!

brandtkeller (approved):
Example in the docs is very helpful. I like that this handles a number of edge cases - more than simply file read as string.

PR #729, opened 2024-10-10
feat(console): add support for multiple input files

### LINKED ISSUE #728: (Console): Support multiple files as input

Currently only a single file can be input into `lula console -f <single-file>`. Need to support multiple files passed to comply to OSCAL standards that each model reside in a separate file. Should also account for writing back to the correct file.

- **Given** multiple files are passed to the lula console
- **When** the console is loaded
- **Then** all files/models are reconciled properly

None

### Additional context

None

### Expected Deliverable

Updated cmd such that `lula console -f ` accepts multiple files

meganwolf0:
## Description

Supports multiple OSCAL files as input as `lula console -f <comma separated list>`

This will merge the same model types (where possible) and create an output file name for any models that will be written (currently only `component` is supported in write, but assuming future models will be added)

Also added an option to specify output file names (again, only component is supported, but as more models get added for "write" capabilities, assuming the flags will grow)

## Related Issue

Fixes #728 

## Type of change

## Checklist before merging


brandtkeller (dismissed):
Generally see the intent here. 

Would you be open to a future issue around the possibility where a write operation doesn't require the creation of a new file? Something about writes not persisting to the original file feels like a gap we can address in the future.

brandtkeller:
Conflict needing resolve otherwise looks good.

brandtkeller (approved):

PR #739, opened 2024-10-15
chore(scanning): create gosec integration in pipelines

### LINKED ISSUE #738: GoSec pipeline integration

Integration of GoSec into pipelines. Resolving the findings.

### Additional context

Reducing our findings that will be resolved by generic scanners.

### Expected Deliverable

- [ ] Addition of an workflow
- [ ] Resolve the existing findings

brandtkeller:
## Description

Integrate [GoSec](https://github.com/securego/gosec) into our pipeline workflows and address findings.

Areas where false positives existed were [annotated](https://github.com/securego/gosec?tab=readme-ov-file#annotating-code)

It may be good to start to document stances around what we allow as far as files that Lula will attempt to operate on and look to provide some guard rails. We do not provide any protections to date on accessing files that should not be accessed. 

I'd also like to consider what revisiting these code annotations looks like. This is natively supported but on what cadence should we conduct a review is TBD. 

Review the [overview](https://github.com/defenseunicorns/lula/security/code-scanning?query=pr%3A739+is%3Aopen) and check open vs closed to see findings resolved. 

## How to Review this PR

Things to note:
- Attempted (in previous commits) to use the github action for gosec
  - Found issues with regards to the exit status reporting failing when there were no issues or results reported
  - Created an issue in the gosec repo - see the reference below
  - Decided to use the binary installation with `go install` which returns the same result as running locally
- Made updates where applicable otherwise used the annotation mentioned above to flag as false positives.
  - References to file paths as variables are expected given our stance
  - where we were implicitly ignoring errors I either added a check (when the resulting function returned an error) or made an explicit ignore and added a small code comment to reference.

## Related Issue

Fixes #738 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/common/composition/composition.go
+++ b/src/pkg/common/composition/composition.go
@@ -42,7 +42,7 @@ func New(opts ...Option) (*Composer, error) {
 
 // ComposeFromPath composes an OSCAL model from a file path
 func (c *Composer) ComposeFromPath(ctx context.Context, path string) (model *oscalTypes_1_1_2.OscalCompleteSchema, err error) {
-	data, err := os.ReadFile(path)
+	data, err := os.ReadFile(path) // #nosec G304
 	if err != nil {
 		return nil, err
 	}
--- a/src/pkg/common/composition/composition.go
+++ b/src/pkg/common/composition/composition.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/defenseunicorns/go-oscal/src/pkg/uuid"
 	"github.com/defenseunicorns/go-oscal/src/pkg/versioning"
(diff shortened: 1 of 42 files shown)


src/pkg/common/composition/composition.go
```diff
@@ -42,7 +42,7 @@ func New(opts ...Option) (*Composer, error) {
 
 // ComposeFromPath composes an OSCAL model from a file path
 func (c *Composer) ComposeFromPath(ctx context.Context, path string) (model *oscalTypes_1_1_2.OscalCompleteSchema, err error) {
-	data, err := os.ReadFile(path)
+	data, err := os.ReadFile(path) // #nosec G304
```
mildwonkey: I would suggesting following the linter suggestion and add a filepath.Clean()  step instead of ignoring the issue here, it is a good suggestion and something I've run into enough times that I should really remember (and yet I don't, so yay linters!) 😅

brandtkeller: Great point - we have a number of locations reading files for which we could implement a clean.

mildwonkey (approved):
LGTM! 
And ty, this reminded me to remove gosec from the other linter-enabling PR so we won't get duplicates. Is this going to be a required check?

brandtkeller:
> LGTM! And ty, this reminded me to remove gosec from the other linter-enabling PR so we won't get duplicates. Is this going to be a required check?

Everything is currently resolved to this point - I will make the check required and we can continue to iterate if it becomes a blocker.

PR #740, opened 2024-10-15
chore: address linter report - swallowed errors and minor test cleanup

mildwonkey:
This PR adds error checking everywhere we were skipping it, and modifies some tests that would return an _error_ if the checked value was nil instead of failing the test, and then go on to inspect the (potentially nil) value. Minor stuff!

There were exceptions where a given method would never return an error; in one case (external method) I added the `nolint` directive, but since I am the only one running that linter right now we could leave it out and just ignore it. In another case I removed the error return, since we never returned an error anyway. 

I'd like to eventually get the golangci-lint linter configured and running in our CI - there are a couple of checks I appreciate, and I see we have cspell so we could get that dictionary updated and running against the whole codebase, etc. I don't have a vision for initial set up though, but this seemed worth doing even in isolation.

I didn't touch anything in the `tui` package since @meganwolf0 has a bunch of work in flight right now!

### DIFF
--- a/src/cmd/common/setup.go
+++ b/src/cmd/common/setup.go
@@ -7,8 +7,7 @@ import (
 	"github.com/defenseunicorns/lula/src/pkg/message"
 )
 
-func SetupClI(logLevel string) error {
-
+func SetupClI(logLevel string) {
 	match := map[string]message.LogLevel{
 		"warn":  message.WarnLevel,
 		"info":  message.InfoLevel,
@@ -37,6 +36,4 @@ func SetupClI(logLevel string) error {
 	}
 
 	printViperConfigUsed()
-
-	return nil
 }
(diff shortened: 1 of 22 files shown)


brandtkeller (commented):
I love the initiative - if this establishes a level of conformance - would it be worthwhile to add a CI check now or do you want to create an issue for follow-on? 

CI execution and baseline threshold for these standardized checks are top of mind for me right now as I work on gosec integration. Likely we have some good overlap 🙂 .

mildwonkey:
I would like to get some linters configured and in the pipeline, yes! This PR doubles as me seeing if there are objections to acting on the default linters. I'm more passively thinking about which linters I find valuable vs. noise. Noise was a big problem at my last place, we added a _lot_ of steps to the build that added time, noise, and annoyance, and did not add people fixing actual issues, so I'm a little hesitant to add anything before I feel like it's *needed*. 

Having said that, swallowed errors have caused me disproportionate amounts of troubleshooting time - worse than a misplaced tab in yaml 😉  so that's one I've become partial to.

brandtkeller (approved):

PR #742, opened 2024-10-16
chore(build): add golangci-lint to the pipeline as a non-blocking step

mildwonkey:
This is a fairly simple golang linter pipeline. It's not required so we don't have to fix anything before merging if we don't want, which will give us some time to tweak the linters (disable linters or individual rules as needed) - like all metrics it's important that we care about and act on the information instead of just add noise. 

I will set myself a weekly reminder for the next 4 weeks to revisit the results during our sync and upgrade this check to required once we're happy with it. I'll keep an eye on the PRs and track what checks are useful vs not and upgrade as we go. 

📢 👋🏻   Let me know what you think of annotating the PR; I don't know if I like it or not. I have strongly mixed feelings on it 😂 Maybe I'll like it more when I'm confident we agree with the linters

---------
# Enabled Linters

I don't think I have a strong opinion on most of these and I'm entirely willing to change the list as many times as we feel like; I like `prealloc` but I'm already worried it'll have too many false positives to be worthwhile. I'm also not sure how seriously I take the `printf: non-constant format string in call` errors but I'll make another pass at fixing them and see how I feel about it afterwards.

Also, while I included gosec in here, we might want to move that into a security specific pipeline. I don't have an opinion there, let me know if you do 🤷🏻 

Defaults:
- errcheck
- gosimple
- govet
- ineffassign
- staticcheck
- unused

Added by me:
- gosec
- gofmt
- goimports
- prealloc


mildwonkey:
I noticed before that the PR Title check doesn't seem to see the change after editing a PR description (even if I re-run the job it was still failing with the original PR description) so I slipped a fix into this PR and added `edit` to the list of triggers.

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+      
+      - name: Setup golang
+        uses: ./.github/actions/golang
+      
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 #v6.1.1
```
brandtkeller: Doubt the space matters - but for uniformity 
```suggestion
        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
```

brandtkeller (changes requested):

brandtkeller (approved):
thank you for doing a better job at naming jobs than I have in the past 😆

Issue #617, opened 2024-08-26
Logging/Messages Clean-up


meganwolf0:
### Is your feature request related to a problem? Please describe.

While digging into our `messages` package a few questions/comments/thoughts arose. Capturing in a question issue for further discussion/dissemination into possible follow-on issues

- Currently we output to stderr - wondering it would better align to best practice to report information to stdout
- We default print to a logfile in a tmp directory. Is this needed/valuable? (It also doesn't appear there's a way to pass a log file unless I'm missing something)
- If we are implementing bubbletea, might be more aligned to try refactor our table-building methods to use tea and our output logs to charm/log
- Should we implement using log/slog as a standard logging library?

### Expected Deliverable

Discussion around questions/comments, follow-on issues where warranted

brandtkeller:
Adding here instead of in a separate issue for the moment. Consideration for ensuring that all logging is done in the `cmd` directories and all internal/public libraries do not use any logging functionality if possible.

Issue #601, opened 2024-08-14
mute validation output in tests


meganwolf0:
### Describe what should be investigated or refactored

When we run validations in testing, there's a whole bunch of data logged to stdout that is just noisy. It would be nice to either implement a log file for tests or implement a "silent" log level so that we could mute that info when running tests.

### Links to any relevant code

None

### Additional context

None

### Expected Deliverable

1) Internal decision on desired solution - might be both implementing/using a log file and/or silent log-level
2) Update `src/messages` and `e2e` tests accordingly

PR #744, opened 2024-10-17
feat(console): assessment results generation

### LINKED ISSUE #652: (Console): Assessment Results generation

Provide the functionality to update the assessment results view+data with the generated information obtained from `lula validate`

- **Given** component definition exists
- **When** user executes some keystokes (e.g., `v` for `validate`)
- **Then** `lula validate` methods execute, add data to the existing model, update the assessment results view

n/a

### Additional context

Where should the key-command live (on the component defn view or assessment results view?). This will require some updates to some of the oscal functions that can write assessment results to the current file. 
(If we put the calling function on the `assessment-results` page it might make more sense to cage as `generate` since it's in the context of the assessment results? Which might be confusing from a CLI perspective, but might be more consistent if we have that nomenclature on every model)

### Expected Deliverable

Determination on user interface/experience

Implementation + Tests

Documentation + gifs

meganwolf0:
## Description

Adds a `lula validate` workflow in the component definition model
* Simple validation preview -> "validating" -> "complete" visuals. Would be nice to provide more feedback to the user in console (esp for longer running validations), but this will probably touch quite a bit of messages/that implementation, so possibly that could be worked into other messaging related issues (#617, #601)
* Updated assessment-results, specifically the implementation of observation props with the validation ID - intent is to draw a thread through to actually showing the validations linked to observations and underlying resources generated as well (hence why I implemented the `RunValidations` method as a wrapper around "ValidateOnControlImplementations" - instead of one of the higher-order methods - so that we could persist the validationstore + underlying validation data for the user to access in assessment results)

## Related Issue

Fixes #652

## Type of change

## Checklist before merging


docs/console/assessment-results.md
```diff
@@ -0,0 +1,40 @@
...
+
+The Assessment Results model responds to the following keys for navigation and interaction (some widgets have additional key response, see respective help views for more information):
+
+| Key | Description |
+|-----|-------------|
+| `?` | Toggle help |
+| `ctrl+c` | Quit |
+| `tab` | Tab right between models |
+| `shift+tab` | Tab left between models |
+| `←/h` | Navigate left across widgets in model|
```
mildwonkey: In the rendered view, these all look like a singlar key combo. I'd leave the `/` (or) outside of the single quotes:

`←`/`h`
 `↓`/`j` 
etc

meganwolf0: updated!

brandtkeller (approved):
Documenting for posterity - approving in current statet with issues to follow around opinionation in the `assessment-results` write process.

Also want to have a general discussion around design of "data context" and how we read/write oscal.

PR #746, opened 2024-10-17
chore: appease the linter

mildwonkey:
This does not need to wait on https://github.com/defenseunicorns/lula/pull/742, but it satisfies MOST of the linting errors that you can see in that PR. I didn't touch the `tui` package yet because I know @meganwolf0 was actively working in that area though I'm happy to add it in here if we'd like, or get to it later.

This was not requested by the linters, but seeing a test setup config file just hanging out there felt weird to me. In hindsight this barely feels less weird so I'm happy to revert if y'all don't like it. 
`renamed:    src/test/e2e/kind-config.yaml -> src/test/e2e/testdata/kind-config.yaml`


brandtkeller (approved):

brandtkeller:
Feedback from #742 helped add context to these updates.  See no reason not to merge this now and evaluate impact to changes in #742.

PR #752, opened 2024-10-21
DRAFT feat(domains): api domain options

mildwonkey:
This draft pull request supplements the (internal) design doc for the API options. No real new functionality is implemented in this PR yet, just refactoring to add the options and show where they might go.

Note for reviewers: Take a look at the new `ApiOpts` and the fields added to the `Request` struct (`ApiOpts` is a field too, any of the APIOpts can be overridden on a per-request basis). The items in ApiOpts have the potential to apply to every API request - for e.g. proxy and timeouts - while the added fields in requests are options that are (most likely to be) configured on a per-API basis. 

This is not complete, I have not added every option from the design doc yet. I was working through them and got thinking about the delineation between client and request options, and here we are.

### DIFF
--- a/src/pkg/domains/api/types.go
+++ b/src/pkg/domains/api/types.go
@@ -42,18 +42,19 @@ type ApiSpec struct {
 
 // Request is a single API request
 type Request struct {
-	Name   string            `json:"name" yaml:"name"`
-	URL    string            `json:"url" yaml:"url"`
-	Method string            `json:"method,omitempty" yaml:"method,omitempty"`
-	Body   string            `json:"body,omitempty" yaml:"body,omitempty"`
-	Params map[string]string `json:"parameters,omitempty" yaml:"parameters,omitempty"`
+	Name         string            `json:"name" yaml:"name"`
+	URL          string            `json:"url" yaml:"url"`
+	Method       string            `json:"method,omitempty" yaml:"method,omitempty"`
+	IsExecutable bool              `json:"executable,omitempty" yaml:"executable,omitempty"`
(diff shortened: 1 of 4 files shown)


src/pkg/domains/api/types.go
```diff
@@ -48,10 +36,25 @@ func (a ApiDomain) IsExecutable() bool {
...
 // Request is a single API request
 type Request struct {
-	Name string `json:"name" yaml:"name"`
-	URL  string `json:"url" yaml:"url"`
+	Name         string            `json:"name" yaml:"name"`
+	URL          string            `json:"url" yaml:"url"`
+	Method       string            `json:"method,omitempty" yaml:"method,omitempty"`
+	IsExecutable bool              `json:"executable,omitempty" yaml:"executable,omitempty"`
+	Body         string            `json:"body,omitempty" yaml:"body,omitempty"`
+	Params       map[string]string `json:"parameters,omitempty" yaml:"parameters,omitempty"`
```
brandtkeller, line 50: do you have an example of `params` use that can help me understand the delineation of them here vs in `ApiOps`?

mildwonkey: Sure! In the design doc `parameters` is described as key-value pairs that Lula will append to the URL (with ? and &); an actual example could be an API query with filters.

brandtkeller: okay! I was looking to understand the delineation for when something might belong in the core `Request` struct vs `ApiOpts` . 

I can see the distinction better after thinking through the separation.

brandtkeller (commented):
Appreciate the state of this draft. Besides questions for my education, I don't have immediate concerns.

Issue #693, opened 2024-09-30
Feature Request: use go-getter/v2 for filedownloads


mildwonkey:
Using [go-getter](https://pkg.go.dev/github.com/hashicorp/go-getter/v2) as the implementation of network.Fetch allows downloading local and network files, as well as a number of other network protocols. It also handles credentials for things like git and s3 downloads. 

Things to consider: go-getter will download zipped archives and directories, not just files (though we can use GetFile vs. GetAny anywhere we want/need to control that - especially for the initial implementation let's not change the behavior for any pre-existing fetching, but I want this in the files domain), so code that assumes there's a 1-1 relationship between the filepath and the files downloaded likely needs to walk the downloaded filepath to get the list of files in the directory.

PR #760, opened 2024-10-24
chore(file domain): add network.DownloadFile and use go-getter to download files

mildwonkey:
## Description
This PR adds a new method, `network.DownloadFile`, which users go-getter to download files. It's currently in use by the file domain. Go-getter's Get supports local files, network downloads (`network.Fetch` also already handles that!) and also a number of protocols such as ftp, s3, git, etc. 

I limited the functionality to downloading INDIVIDUAL FILES and not full directories and repositories _for now_, since it would take more testing and a slightly larger refactor of the file domain. (I have an idea for that, but wanted to get this PR dealt with first). A future PR will add `network.Download` for directories, git repos, etc. (This will require an additional field in the file spec, `source` or something similar. Users will still need to list the files that the validations run against).

I added one test that pulls a file from our git repo to the e2e tests. There is a comment in the file warning future-us of that, but it's still worth mentioning that will break and annoy us some day (and be easily fixed, but still. Annoying). 

There are some other changes in the e2e test files - It was getting hard to find the actual test output, so I removed all the calls to message.Info. I also noticed I had randomly started using assert where I wanted require in the files domain e2e tests, so that's fixed too 🙄 

...

## Related Issue
Relates to #693 

## Type of change

more of a precursor to a new feature, but close enough?

### DIFF
--- a/src/pkg/domains/files/files.go
+++ b/src/pkg/domains/files/files.go
@@ -68,7 +68,6 @@ func (d Domain) GetResources(ctx context.Context) (types.DomainResources, error)
 		if err != nil {
 			// Assign empty data value for reporting purposes
 			tmpDRs[fi.Name] = map[string]interface{}{}
-			filenames[file] = fi.Name
 			errs = errors.Join(errs, fmt.Errorf("error writing local files: %w", err))
 			continue
 		}
--- a/src/pkg/domains/files/files.go
+++ b/src/pkg/domains/files/files.go
@@ -8,9 +8,10 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/open-policy-agent/conftest/parser"
+
 	"github.com/defenseunicorns/lula/src/pkg/common/network"
 	"github.com/defenseunicorns/lula/src/types"
-	"github.com/open-policy-agent/conftest/parser"
 )
 
 type Domain struct {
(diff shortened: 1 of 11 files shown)


meganwolf0:
Is the intent to completely move off network.Fetch, or is this supposed to just be like a separate functionality with the intent to support more specialized downloads specifically for the file domain?

Also, because I was just at a security conference and it's fresh, looking at like the dependencies for the current network package (which are all go built-ins) vs this PR which adds some super old/forked package deps (I'm sure we have a ton of that, honestly this is really the first time I'm really looking under the hood there), just thinking more about our posture on these kinds of things/potentially accepting risks. More of a team discussion I think though.

---
Edit after thinking about the security comment a bit more -
Maybe an alternative solution would be to incorporate the go-getter into the `files` domain instead of `pkg`, because it definitely has the capacity to support better use cases than just simple file download, but also maintain network.Fetch in our `pkg`/core Lula features since that is really only used for compose and I don't really see much benefit to swapping to go getter (maybe there is, I suppose for the auth case if we have to auth to compose... sidebar since that's also a can of worms with specifying where that data would be). Then if we wanted to like more discretely break out domains in the future (for like security things), we wouldn't be strictly dependent on go getter for core lula functions.

mildwonkey:
This isn't going to replace Fetch, since go-getter only writes files and most callers just need to read the bytes. My original files domain PR just used go-getter in the domain and I don't mind doing that again.

No argument on the general dependency concern, but if we make the domains external plugins, go-getter is the library I'd suggest we use to download them. It is indeed always a risk, but at least go-getter has a ton of attention on it thanks to several projects, instead of managing it all myself or trusting a library I've never heard of with 6 stars (side note, I'd love an approved go libraries list). Having said that, if we don't feel good about this, I have no problem closing it till we can discuss more.

meganwolf0:
> This isn't going to replace Fetch

Ok great, I was thinking the original issue sort of indicated that but since not then sounds good to me!

> at least go-getter has a ton of attention on it thanks to several projects

Agreed, seems like there's a few big projects importing the library so not super concerned. I think just fresh off that conference it was more in the forefront of my mind that probably goes into the checklist for PR reviews.

src/test/e2e/file_validation_test.go
```diff
@@ -120,8 +120,16 @@ func TestFileValidation(t *testing.T) {
...
+	})
+
+	// This test fixture is referencing a file on GIT, so if you're moving
+	// things around here you should probably check that, too
+	t.Run("remote file download", func(t *testing.T) {
+		ctx := context.WithValue(context.Background(), types.LulaValidationWorkDir, passDir)
+		validator, err := validation.New()
+		require.NoError(t, err)
+		_, err = validator.ValidateOnPath(ctx, "scenarios/file-validations/pass/component-definition-remote-files.yaml", "")
+		require.NoError(t, err)
```
meganwolf0, line 127: should we checked that the result was satisfied? The validations will fail gracefully even if a file couldn't be downloaded, so probably good to check that it passed because that'll actually be the check of the remote download

meganwolf0 (changes requested):

src/pkg/domains/files/files.go
```diff
@@ -131,13 +134,22 @@ func (d Domain) GetResources(ctx context.Context) (types.DomainResources, error)
 
 	// add the string form of the unstructured files
 	for _, f := range unstructuredFiles {
-		// we don't need to copy these files, we'll just slurp the contents into
-		// a string and append that as one big DomainResource
-		path := filepath.Clean(filepath.Join(workDir, f.Path))
-		b, err := os.ReadFile(path)
+		stringdir, err := os.MkdirTemp(dst, "string")
```
brandtkeller, line 155: Do we need a temporary directory _for each_ file or would one suffice?

brandtkeller: there may be an underlying reason for this - primarily just curious in the event we were to expose an option for a file cache on the filesystem.

mildwonkey: This is a temp dir for all files by user-specified parser, not necessarily every individual file.

brandtkeller: maybe my brain is broken tonight - we have this slice of unstructured files (string parser)- we iterate over each file and we create a temporary directory before performing the copy/read. 

Doesn't each loop create a new temporary directory?

mildwonkey: No - in this code section you can look at the `dst` line: it's downloading just the file into the `stringdir`. It's not creating a directory for each file. All the unstructured files go to the "string" parser directory. 

Go-getter in file download mode accepts a full file path, not a directory, to install files into.

brandtkeller (approved):

mildwonkey:
@meganwolf0 - i did push the extra test validation you requested, does it look sufficient to you?

meganwolf0 (approved):

PR #762, opened 2024-10-24
feat(dev): observation print commands

### LINKED ISSUE #621: Dev commands to extract validation and resources by observation uuid

Create some dev-only commands that ingest the observation ID and report on the related validation and optional resources (if saved)

- **Given** resources are saved from a validation
- **When** `lula dev print-resources -o <observation_uuid>`
- **Then** resources used are returned

none

### Additional context

This feature is mainly geared at enabling advanced debugging of failing validations, particularly in CI environments. Assuming these commands might be hidden/useful mostly for Lula development/debugging purposes.

### Expected Deliverable

Additional dev commands + Associated documentation

meganwolf0:
## Description

Dev commands to print the associated resources and validation provided an Observation UUID
* Primary intent is to more easily identify associated resources/validation from observations when they don't pass - additional context is this was initially an effort for dash days to have automation surrounding previously passing, now failing observations/validations
* Limitation on the `print-validation` command is that the component definition must be composed, so the ID can be referenced from the back-matter. This is probably a larger can of worms on supporting the file path in that prop, but I feel like right now the cleanest case is to enforce that it must be composed

## Related Issue

Fixes #621 

## Type of change

## Checklist before merging


src/cmd/dev/print-resources.go
```diff
@@ -0,0 +1,130 @@
...
+var printResourcesHelp = `
+To print resources from lula validation manifest:
+	lula dev print-resources --assessment /path/to/assessment.yaml --observation-uuid <observation-uuid>
+
+To print resources from lula validation manifest to output file:
+	lula dev print-resources --assessment /path/to/assessment.yaml --observation-uuid <observation-uuid> --output-file /path/to/output.json
+`
+
+var printResourcesCmdLong = `
+Print out the the JSON resources input that were provided to a Lula Validation, as identified by a given observation and assessment results file.
```
mildwonkey: duplicate `the`

mildwonkey:
Seeing as we're re-using the `print` suffix, what do you think about `lula dev print` as a subcommand that takes either flags or arguments to identify what it's printing? Ie `lula dev print --validation=<>`, or `lula dev print resource`? (as a future feature we could then allow multiple print statements in one command)

meganwolf0:
> Seeing as we're re-using the `print` suffix, what do you think about `lula dev print` as a subcommand that takes either flags or arguments to identify what it's printing? Ie `lula dev print --validation=<>`, or `lula dev print resource`? (as a future feature we could then allow multiple print statements in one command)

Yeah looking back at the commands, there's a ton of logic overlap so that might be a fair way to refactor that. Let me poke at that a bit more and throw this back up

meganwolf0:
@mildwonkey it's been thrown up 🤮 🤣

mildwonkey:
I like it!! I think you still need to update the lula_dev docs but otherwise 👩🏻‍🍳 💋

brandtkeller (commented):
The logic itself I don't have much in the way of concerns. I'd like to document the constraints around composed component definitions as that looks like an obvious "gotcha" that will occur. 

Otherwise my primary concern is command location - does a sub-command of `dev` feel appropriate here if the story we've been telling to-date is that `dev` is meant for `validations` and separate/distinct from OSCAL?

meganwolf0:
> The logic itself I don't have much in the way of concerns. I'd like to document the constraints around composed component definitions as that looks like an obvious "gotcha" that will occur.

<strike>Sounds good, I'll update the command help to try and highlight that constraint more. Might also add like a message to the user if a validation isn't found that the user should check they're using the `composed` component defn.</strike>

Actually, I just added the compose to that step - the wrinkle to that is that it doesn't _currently_ support templating, but that might be something we can add via #771 ?

> Otherwise my primary concern is command location - does a sub-command of dev feel appropriate here if the story we've been telling to-date is that dev is meant for validations and separate/distinct from OSCAL?

I guess I wasn't thinking `dev` was explicitly for validations, although I suppose that is all the functions at the moment, I think I was considering it more like as developer helper functions... but could do a different top level command for these? Or could just make the `print` it's own top level thing.

Our help text for dev is "Collection of dev commands to make dev life easier", our help text for tools is "Collection of additional commands to make OSCAL easier"... feels like this is a little of both haha, although print logic is more lula <> OSCAL specific rather than standalone OSCAL (since we're relaying on specific props/links) so maybe tools doesn't make sense either.

brandtkeller:
> Actually, I just added the compose to that step - the wrinkle to that is that it doesn't currently support templating, but that might be something we can add via https://github.com/defenseunicorns/lula/issues/771 ?

That sounds good. 

> Our help text for dev is "Collection of dev commands to make dev life easier", our help text for tools is "Collection of additional commands to make OSCAL easier"... feels like this is a little of both haha, although print logic is more lula <> OSCAL specific rather than standalone OSCAL (since we're relaying on specific props/links) so maybe tools doesn't make sense either.

My wanting to think about command locality is still an open question that I've made for if `lula dev` could be used to support validation workflows without OSCAL - highlighting the value of the validations in isolation. 

`lula tools` is typically where we have been dropping OSCAL helpers outside of the core functionality. That could be a contender? Optimally the Logic having a primary focus on OSCAL means that it would operate on OSCAL generically and lacking the required information would not outright error. 


(maybe not in scope today) - This highlights another area of opportunity which is documenting the design - as this may be Lula opinionation today - but something we should look to have available for review and standardization from the OSCAL community provided there was interest - either in changing our logic or others following the pattern.

meganwolf0:
@brandtkeller went ahead and moved the commands to `tools`, lmk if you think that fits better

brandtkeller (commented):
Minor conflict to resolve

meganwolf0:
Super minor change - moved the `WriteResources` function down to `src/types` as that felt like a better fit... after getting deep in the validation tests/this dir in particular, it feels like a good step toward a slow disaggregation of that `types.go` into more relevant names :clarity-is-kindness:

mildwonkey (approved):
I like the organizational changes!

brandtkeller (approved):

PR #766, opened 2024-10-28
feat(api domain): extend options for Get requests

### LINKED ISSUE #765: Extend Get options
breaking #356 into smaller PRs. the first will focus on the options that are relevant to GET.

mildwonkey:
## Description
Extends the optional configuration for API Domain Get methods: 
- parameters (alternative to supplying in the URL)
- Headers
- Timeout
- Proxy configuration

I wanted to work through the PR review before I spend too much time on documentation so I opened separate docs tickets for each phase (I split up the main API Domain Optionality ticket by HTTP method to keep the PRs smallish).

Some notes of interest to reviews:
- I added a validation & mutation step that reads in the APISpec and normalizes some values, such as the timeout string to a time.Duration, and stores the normalized values in unexported fields in the APISpec struct. I then made GetResources a method on the ApiDomain so I didn't have to pass the Spec around between methods. 
- Right now I'm treating ApiOptions in the `Request` as a full override of any top-level api options. I'm open to adding a flag to make that a merging behavior but wanted feedback before adding that (user or 🦄)

## Related Issue

Fixes #765 (documentation is a separate PR)

## Type of change

## Checklist before merging


src/pkg/common/types.go
```diff
@@ -153,7 +154,8 @@ func (validation *Validation) ToLulaValidation(uuid string) (lulaValidation type
 	domain, err := GetDomain(validation.Domain)
 	if domain == nil {
 		return lulaValidation, fmt.Errorf("%w: %s", ErrInvalidDomain, validation.Domain.Type)
-	} else if err != nil {
+	}
+	if err != nil {
```
mildwonkey, line 158: 👋🏻 Note for reviewers here! It took me <REDACTED> to figure out that Lula was swallowing an error returned from GetDomain because we only check the error if the returned domain is nil, and the api domain wasn't doing that. I extended this to always check for an error, let me know what you think

meganwolf0: Yeah that seems fair, and looking at this again I'm trying to remember the logic behind why that weird structured check was there - I suppose you could make a case that in the future if domain is nil we could continue if the validation was like a "placeholder" or something... (so we probably would want an if else err) but since that's not implemented, and quite possibly might be better as a separate function, I think what you did there is good.

src/pkg/domains/api/api.go
```diff
@@ -1,57 +1,48 @@
...
+	default:
+		collection := make(map[string]interface{}, 0)
+
+		// defaultOpts apply to all requests, but may be overridden by adding an
+		// options block to an individual request.
+		var defaultOpts *ApiOpts
+		if a.Spec.Options == nil {
+			// This isn't likely to be nil in real usage, since CreateApiDomain
+			// parses and mutates specs.
+			defaultOpts = new(ApiOpts)
```
meganwolf0, line 23: I know this is probably never going to be hit, but for completeness, should we add the default timeout here? I think if these default opts get used that is left as 0 (unless I'm missing somewhere else it's set)

src/pkg/domains/api/http_request.go
```diff
@@ -0,0 +1,73 @@
...
+	q := url.Query()
+
+	for k, v := range queryParameters {
+		// using Add instead of set incase the input URL already had a query encoded
+		q.Add(k, strings.Join(v, ","))
+	}
+	// set the query to the encoded parameters
+	url.RawQuery = q.Encode()
+
+	req, err := http.NewRequest(http.MethodGet, url.String(), nil)
```
meganwolf0: (genuine learning question/comment) would it make sense to do the http.NewRequestWithContext() to pass through the context if it's canceled? I see in `api.go` that the context cancellation is handled but wondering if that flows down to the request process itself?

src/pkg/domains/api/http_request.go
```diff
@@ -0,0 +1,73 @@
...
+		return respTy, fmt.Errorf("error: calling %s returned empty response", url.Redacted())
+	}
+	defer res.Body.Close()
+
+	responseData, err := io.ReadAll(res.Body)
+	if err != nil {
+		return respTy, err
+	}
+
+	if res.StatusCode != http.StatusOK {
```
meganwolf0, line 49: Not sure if there's a valid use case, but wondering if it might ever be useful to not return an error but simply report the status code and response body as the resources of the domain? Or in the request include an expected status code. Like if a user actually wanted to prove that trying to access the API was unsuccessful without authentication or something for example. This might be a future thing to consider, just throwing it out for additional thoughts.

src/test/e2e/api_validation_test.go
```diff
@@ -203,3 +211,73 @@ func TestApiValidation(t *testing.T) {
 
 	testEnv.Test(t, featureTrueValidation, featureFalseValidation)
 }
+
+// TestApiValidation_templated uses a URL parameter to control the return response from the API.
+func TestApiValidation_templated(t *testing.T) {
```
meganwolf0, line 216: This is a snazzy test set-up, that is all.

src/pkg/domains/api/http_request.go
mildwonkey: Agreed! That's in the design doc as a feature and part of why I split this into multiple PRs - I wanted to limit how many new things I was introducing at once. It's definitely planned!

meganwolf0: oops yes you definitely do have that in the design doc 😅 I promise I read it!!

src/pkg/domains/api/api.go
mildwonkey: That's a good point! I think when I originally did this, I wasn't setting a default timeout at all (Lula did not set a timeout previously, so it was preserving the original behavior), and since this situation can only occur in tests I was fine with zero values, but I can add it back in - it's better defensive programming anyway.

src/pkg/domains/api/http_request.go
mildwonkey: Yeah, that's a fair question. I had hoped this would be sufficient because I was having second thoughts about sending the original context with the http request, but there's no reason not to (and I could make a new context if that was an actual concern).

mildwonkey: fixed!

src/pkg/domains/api/http_request.go
mildwonkey: lol it's absolutely fine I'm trying not to sound like a jerk while saying RTFM and ideas are ALWAYS welcome - I think it's a great suggestion and we're both brilliant for having the idea ;)

meganwolf0 (commented):
(This was originally a comment about how something didn't work as expected, but then I realized I had the wrong spec structure 🤦‍♀️ )
Anyway, did some local testing and everything worked swimmingly!

meganwolf0 (dismissed):

src/pkg/domains/api/api.go
```diff
@@ -1,57 +1,49 @@
...
+			} else {
+				client := clientFromOpts(request.Options)
+				responseType, err = doHTTPReq(ctx, client, *request.reqURL, request.Options.Headers, request.reqParameters, responseType)
 			}
-			prettyJson := prettyBuff.String()
-
-			var tempData interface{}
-			err = json.Unmarshal([]byte(prettyJson), &tempData)
 			if err != nil {
 				return nil, err
```
brandtkeller: Following the pattern I recently introduced for collecting what is available for evidence - would it be possible to return what has been collected thus far? 

I imagine (but may be assuming incorrectly) that we would actually want to return on error and not continue with a list of requests if one fails. But returning the data from all previously no error requests may be valuable in reproducing an error?

mildwonkey: Yep, I had planned on introducing that in a future PR since I wasn't sure if your change had landed yet - I'll add the behavior (regardless of _error_) the same time I refactor so Lula will return results whatever the response code.

mildwonkey: 👍🏻 updated it so we're returning a partial collection (domain resources) on error - there will be more in the next PR, where we'll start returning the response regardless of code

meganwolf0: Just as a consistency check, I think the mods made in #704 also return the key, this looks like it does not return the key if the request failed

mildwonkey: It's going to happen in my next PR anyway so I'd rather not change this one further unless you are specifically requesting that I do so now - his PR isn't merged yet so I wasn't feeling rushed

meganwolf0 (approved):

brandtkeller (approved):

PR #777, opened 2024-11-02
feat(dev): allow templating validations in dev workflows

### LINKED ISSUE #771: Add templating to dev commands

Currently `dev validate` and `dev get-resources` doesn't support templating the validation.

- **Given** dev <cmd>
- **When** target is a template
- **Then** target is templated/executed

Run tools template -> then dev <cmd> on that. This is a stopgap solution but the extra step is mildly annoying

### Expected Deliverable

- [ ] Dev validate supports template configuration + --set to override
- [ ] Dev get-resources supports template configuration + --set to override

brandtkeller:
## Description

Add templating with `all` render type to dev commands.

Add command tests for the dev commands. Retain the kubernetes tests while we revise the kubernetes testing plan to utilize CLI. 

The primary content change for this PR is the creation of a `--set` flag on the `lula dev`command and then templating validations. Took this time to update the command structure to match the new standard. 

## Related Issue

Fixes #771 

## Type of change

## Checklist before merging


meganwolf0:
this is totally a nit, but should all the unexported variables in the dev commands be switched to lowercase to be consistent?

meganwolf0:
Also possibly not a totally breaking comment but just convention question - I had been calling most of the comparison files <whatever>.golden. I see that this PR deviates from that. Just wondering if that's a convention we'd like to keep, simply to denote that these are generated files/shouldn't be edited? Alternatively we could do like <file>_golden.<ext>, or we could add additional subdirectories (input v golden) to just delineate more?

brandtkeller:
> Also possibly not a totally breaking comment but just convention question - I had been calling most of the comparison files .golden. I see that this PR deviates from that. Just wondering if that's a convention we'd like to keep, simply to denote that these are generated files/shouldn't be edited? Alternatively we could do like _golden., or we could add additional subdirectories (input v golden) to just delineate more?

Do we need more delineation between inputs vs outputs? if so then I don't have anything against your request. 

I believe I still retain the use of `.golden` where applicable and then otherwise added other testing data to the same directory. If we need to move that out then I can do that as well.

brandtkeller:
> this is totally a nit, but should all the unexported variables in the dev commands be switched to lowercase to be consistent?

Appropriate nit! thanks for the callout. Updated.

meganwolf0:
> Do we need more delineation between inputs vs outputs? if so then I don't have anything against your request

I don't think it's a _need_ per se, more just like clarity on which files are used for what. I think it would just fall into the bucket of easier to see what's what on first glance. And also just like a "you should never directly modify this file" sort of indicator

meganwolf0 (approved):

PR #780, opened 2024-11-05
fix(generate): template nested prose with params

### LINKED ISSUE #779: Param replacement during generation from catalog prose
### Steps to reproduce
1. `lula generate component --catalog-source https://raw.githubusercontent.com/GSA/fedramp-automation/93ca0e20ff5e54fc04140613476fba80f08e3c7d/dist/content/rev5/baselines/json/FedRAMP_rev5_HIGH-baseline-resolved-profile_catalog.json --component 'Istio Controlplane' --requirements ac-14,ac-4,ac-4.21,ac-4.4,ac-6.3,ac-6.9,au-12,au-2,au-3,au-3.1,cm-5,sc-10,sc-13,sc-23,sc-3,sc-39,sc-4,sc-7.20,sc-7.21,sc-7.4,sc-7.8,sc-8,sc-8.1,sc-8.2 --remarks statement --framework il4 -o component.yaml -l debug`
2. Review Control `ac-4.4` prose
3. Should contain " {{ insert: param, ac-04.04_odp.03 }} "
4. This will error during `lula validate` due to lula attempting to apply templating against it

### Expected result

Lula validate can execute

### Actual Result

Error during lula validation

### Visual Proof (screenshots, videos, text, etc)

https://repo1.dso.mil/big-bang/product/packages/istio-controlplane/-/blob/main/oscal-component.yaml?ref_type=heads#L1142

### Severity/Priority

urgent

brandtkeller:
## Description

Adds a `nested` parameter to the current `replaceParams` logic such that we can apply OSCAL prose templating as is used in the OSCAL community against nested items when parameter use is used within a `choice` selection. 

## Related Issue

Fixes #779 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/common/oscal/component.go
+++ b/src/pkg/common/oscal/component.go
@@ -424,7 +424,7 @@ func ControlToImplementedRequirement(control *oscalTypes_1_1_2.Control, targetRe
 			if contains(targetRemarks, part.Name) {
 				controlDescription += fmt.Sprintf("%s:\n", strings.ToTitle(part.Name))
 				if part.Prose != "" && strings.Contains(part.Prose, "{{ insert: param,") {
-					controlDescription += replaceParams(part.Prose, paramMap)
+					controlDescription += replaceParams(part.Prose, paramMap, false)
 				} else {
 					controlDescription += part.Prose
 				}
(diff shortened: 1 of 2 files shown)


src/pkg/common/oscal/component.go
```diff
@@ -616,15 +616,30 @@ func addPart(part *[]oscalTypes_1_1_2.Part, paramMap map[string]parameter, level
 	return result
 }
 
-func replaceParams(input string, params map[string]parameter) string {
+func replaceParams(input string, params map[string]parameter, nested bool) string {
```
mildwonkey, line 619: Can you add test coverage for `replaceParams` (with bools)?

brandtkeller: Feels like cheating but notwithstanding non-exported functionality tests - I updated existing tests for the exported logic to include a known test-case for where this nesting occurs and ensures it is no longer present in any remarks.

meganwolf0:
I don't see any technical issues here, more of a question about the stance where we're just removing all the data inside the {{ }} - wondering if instead we should remove the double curly braces but leave behind the info? Like when I look at ac-4.4 specifically, that insert being removed seems to add additional information about what a user needs to provide (e.g., the final choice of "organization-defined procedure or method"). Kinda a lift to actually template that (though we maybe could) but feels like scrubbing the info completely is overkill

meganwolf0 (approved):

PR #795, opened 2024-11-12
fix: lula validations in store don't persist validation state

### LINKED ISSUE #789: Lula Validations in store don't persist validation state
### Environment
Device and OS:  
App version:    
Kubernetes distro:  
Kubernetes version:  
provider:
- [ ] kyverno
- [ ] opa

### Steps to reproduce
Within a single execution of Lula Validate, a single validation that's linked to multiple controls/targets shouldn't run multiple times
For a single result output this is the behavior, however for multiple results generated, validations are re-executed.

This is a problem both with latency and when trying to access a validation's state outside of executing the validation (e.g., when running tests/getting resources back from an already run validation).

### Additional Context

Really the validation references just need to be switched to pointers

meganwolf0:
## Description

Updates the `RunValidations` function in the validation-store to use pointers to validations to store data across executions -
solves a couple of issues:
- Domain resources data was not able to be accessed - which prevented tests from being run on validations
- The same validations would re-execute across different targets. I believe this should not happen within a single `lula validate`, but if this does break some existing conceptual assumptions around executing Lula Validations, then this might be a larger conversation.

## Related Issue

Fixes #789 

## Type of change

## Checklist before merging

- [] [Schema Updates](https://github.com/defenseunicorns/lula/blob/main/docs/community-and-contribution/schema-updates.md) applied


brandtkeller (approved):
This makes sense - thank you for both identifying the issue and fixing it!

PR #802, opened 2024-11-15
feat(system-security-plan): initial creation of SSP Model

### LINKED ISSUE #797: Create OSCALModel for SSP

Create the OSCALModel interface instantiation for the SSP model

### Additional context

Following the Profile model structure set forward

### Expected Deliverable

- [ ] Library functionality for SSP that is conformant to the OSCALModel interface
- [ ] Associated unit tests
- [ ] Follow-on issues scoping the inclusion of cmd, additional oscal data, etc.

meganwolf0:
## Description

Implements the OSCALModel interface on an SSP.
- Simply the library functions for this PR, follow on addressing the cmd aspect
- Additional data is going to be required to fully complete the SSP, in the meantime left "<TODO...>" remarks to try and indicate additional vetting needed on those fields

## Related Issue

Fixes #797 

## Type of change

## Checklist before merging


meganwolf0:
To discuss:
Are we merging validations (via backmatter) into the SSP? Or are we maintaining the original links to the validations and performing a relative path mapping (to get those links properly added to the ImplementedRequirements.ByComponents? IF we want to go that route, we need to re-tool part of the compose functionality to partially compose compdefs (e.g., get the imported ones, re-write the validation/ref relative paths)

brandtkeller:
> To discuss: Are we merging validations (via backmatter) into the SSP? Or are we maintaining the original links to the validations and performing a relative path mapping (to get those links properly added to the ImplementedRequirements.ByComponents? IF we want to go that route, we need to re-tool part of the compose functionality to partially compose compdefs (e.g., get the imported ones, re-write the validation/ref relative paths)

I do think we should discuss this! Do you see this as a blocker for this PR or simply something we should discuss ASAP (next week)?

meganwolf0:
> Do you see this as a blocker for this PR or simply something we should discuss

Yeah mostly just something to discuss - I don't think it's blocking anything here/whatever decision wouldn't invalidate anything done so far.

src/pkg/common/oscal/system-security-plan.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// GenerateSystemSecurityPlan generates an OSCALModel System Security Plan.
+// Command is the command that was used to generate the SSP.
+// Source is the catalog source url that should be extracted from the component definition.
+// Compdef is the partially* composed component definition and all merged component-definitions.
+// TODOs: implement *partially = just imported component-definitions, remapped validation links;
+// implement system-characteristics, parties->users->components, component status, (probably more);
+// support for target instead of source?
+func GenerateSystemSecurityPlan(command string, source string, compdef *oscalTypes.ComponentDefinition) (*SystemSecurityPlan, error) {
```
brandtkeller, line 124: Doesn't need to block merge - but I would be interested in thinking about common patterns we're performing here if we think about what makes this a potentially good library (or potentially back-porting to go-oscal).  

I don't believe the "compose" functionality is a common pattern. Which makes me think that maybe we should expect a slice of component-definitions here instead of an expectation of having performed some aggregation previously.

brandtkeller (dismissed):
No issues with the logic out-front of more implementation. 

I used the [go-oscal types](https://github.com/defenseunicorns/go-oscal/blob/main/src/types/oscal-1-1-2/types.go#L95) to look at required fields at the top-level object. Reviewed what the logic performed for generation as a primary indicator of the logic. 

I appreciate the consideration that went into the merge functionality. The checks establish some precedence for what is permitted for merge - will be good to document this when the implementation layer is available. 

If possible I would like to see more guidance in PR's for _how_ to review these changes. We're navigating new (to us) models and discovering how each should respond to specific actions. We're on a journey together here to learn how these models work and how to automate them.

src/pkg/common/oscal/system-security-plan.go
meganwolf0: Yeah, that's a fair point. So would you have the 'partial compose' (i.e., just pulling the imported compdefs) inside this function as well? I was kind of trying to simplify this at a first cut (e.g., assuming my compdef was exactly what I wanted) but that logic I'd think could probably be implemented in a follow-on issue. (probably coupled to the "to discuss" point above)

meganwolf0:
> If possible I would like to see more guidance in PR's for _how_ to review these changes

Agreed! I felt similarly when I was recently reviewing the profile model implementation. I can start a working doc somewhere, maybe Notion for now as we work through a process?

src/pkg/common/oscal/system-security-plan.go
brandtkeller: > So would you have the 'partial compose' (i.e., just pulling the imported compdefs) inside this function as well?

I would like to say yes - I am curious if there is any expectation of what this looks like - Is there any documented stance on the relationship between compdef <-> ssp that we could investigate. 

I agree with follow on work - just wanted to call it out so we can start to delineate between Lula opinions and standard OSCAL processes.

brandtkeller:
> > If possible I would like to see more guidance in PR's for _how_ to review these changes
> 
> Agreed! I felt similarly when I was recently reviewing the profile model implementation. I can start a working doc somewhere, maybe Notion for now as we work through a process?

Open to anything that helps spread awareness. Some of this can be self-documenting but my request is simply some information in the PR that describes or establishes context for the changes. 

This PR touches the SSP OSCAL model - here is associated information to review. This work follows 'x' pattern and you can see linked documentation support the logic. Lula applies opinionation to the use of 'y' field in the SSP type for 'z' reason. etc etc

src/pkg/common/oscal/component.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/defenseunicorns/go-oscal/src/pkg/uuid"
 	oscalTypes_1_1_2 "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-2"
+
 	"github.com/defenseunicorns/lula/src/pkg/message"
```
mildwonkey: This is, unfortunately, exactly the annoying thing goimports does sometimes - you need to manually remove the extra newline here before it'll properly sort.

CloudBeard (commented):
This looks correct taking what we have with Component Definitions and getting to SSP.

The relationship between CompDef < > SSP doc could be created separate of this PR. I think that doc with deeper/refresh research from last quarter would bring up some optional prop items we could add to CompDefs and start the opinionation route.

brandtkeller (approved):
Approving and resolving current conversation. Believe this is setup well with follow on work identified and considerations we might be able to pull in.

PR #818, opened 2024-11-21
feat(profile): resolve all controls in profile

### LINKED ISSUE #813: Get all controls from imported profile

Given a provided profile model, we need functionality to retrieve all controls (that are specified from the `profile.imports`)

- **Given** OSCAL Profile model
- **When** method is applied to model, e.g, getAllControls
- **Then** the imports are recursively searched (can contain both catalog or other imported profiles) for all unique controls

None

### Additional context

This is required to fully compile the SSP from a profile input.

### Expected Deliverable

- [ ] Updated `src/pkg/common/oscal/profile` model methods + unit tests

meganwolf0:
## Description

Library functions to resolve all controls from a provided profile. Returns a map of source UUIDs to the respective controls. Intent of this structure is to properly map referenced source (from comp-defn) to the respective controls that are dictated by the profile.

Important Additions/Modifications:
* Refactored the extraction of controls from a provided profile in `component.go` - the addition of this PR should allow for generation of component directly from a profile, HOWEVER, note the below constraint on profiles using internal refs
* Updated network functionality to read from local files - should handle multiple path types, I think it was limited before on "file://" definitions which aren't always picked up for local paths
* Unit tests for resolving catalog and profile controls

Important Note - 
Including profiles with references to internal UUIDs (e.g., the way NIST does [this profile](https://github.com/usnistgov/oscal-content/blob/main/nist.gov/SP800-53/rev5/yaml/NIST_SP-800-53_rev5_MODERATE-baseline_profile.yaml)) will not resolve.

## Related Issue

Fixes #813 

## Type of change

## Checklist before merging


src/pkg/common/network/network.go
```diff
@@ -163,7 +163,12 @@ func GetLocalFileDir(inputURL, baseDir string) string {
 	}
 	requestUri := url.RequestURI()
 
-	if url.Scheme == "file" {
+	// Intent of check is to handle different specifications of file paths:
+	// - file:///path/to/file
+	// - ./path/to/file
+	// - /path/to/file
+	// - https://example.com/path/to/file
+	if url.Scheme == "file" || !url.IsAbs() {
```
brandtkeller, line 171: Would this imply that people could start adopting paths in the OSCAL that might violate OSCAL [URI Usage](https://pages.nist.gov/OSCAL/resources/concepts/uri-use/)?

brandtkeller: The docs are ambiguous - but I am curious if the `scheme` is a required part of the URI use in OSCAL.

meganwolf0: I don't think `file` prefix is required - if you look at some of the [oscal examples](https://github.com/usnistgov/oscal-content/blob/main/examples/component-definition/yaml/example-component.yaml) for fields that specify uri (see href in backmatter) those just do relative paths

meganwolf0:
Temporarily pulling this down to address a bug I found in component generate - will also update those tests accordingly.

meganwolf0:
> Temporarily pulling this down to address a bug I found in component generate - will also update those tests accordingly.

Realized that the `AddControl` logic was breaking the control search loop before child controls could be checked, so needed to re-order some of that functionality. Also updated the debugging logs in `ComponentFromCatalog` to accurately reflect controls not-found. Updated the `TestComponentFromCatalog` to check for an additional control that is nested under an excluded control.

src/pkg/common/network/network.go
brandtkeller: Roger! would be good to validate this with the constraints we talked about in metaschema and oscal-cli but otherwise sounds good.

brandtkeller (approved):
Great point of iteration - I don't see anything that has to block progress but this does generate a lot of fun opportunities.

PR #826, opened 2024-12-02
feat(opa): allow importing custom OPA modules in OPA policies

nacx:
## Description

This change allows importing existing OPA packages to be used in the main OPA policies.

In scenarios where several controls are implemented, it is desirable to have a way to define common functions that can be reused across them, to keep the main policies understandable and maintainable. OPA already provides a nice way of packaging modules, and this PR adds a `modules` configuration knob to the OPA provider to allow importing existing rego modules.

## Related Issue

N/A

## Type of change

## Checklist before merging

- [x] Test, docs, are added or updated as needed


src/pkg/providers/kyverno/types.go
```diff
@@ -9,14 +9,11 @@ import (
 )
 
 type KyvernoProvider struct {
-	// Context is the context that the Kyverno policy is being evaluated in
-	Context context.Context `json:"context" yaml:"context"`
-
```
nacx, line 14: Note that I've removed the Context from the Provider objects. They were mostly used to carry the context that was passed to the Provider constructor and make it available in the evaluation phase.
However, there is no good way (yet) to pass that context, as the provider creation is called from many different points when the validation model is built (and that's probably the reason for [this comment](https://github.com/defenseunicorns/lula/blob/main/src/pkg/common/types.go#L152-L154)), and it still feels incorrect to use that context for the evaluation phase.

The evaluation phase is triggered from different code paths that already carry a context, and that is the one that should probably be propagated. To that end, later in this PR I modify the `Provider.Evaluate` method to accept the context. This way the evaluation gets the right context instead of always receiving a `context.Background()`.

src/pkg/providers/opa/opa_test.go
```diff
@@ -0,0 +1,89 @@
...
+				Rego:    "package validate\n\ndefault validate = true",
+				Modules: map[string]string{"empty": "testdata/empty.rego"},
+			},
+			wantErr: opa.ErrCompileRego,
+		},
+		{
+			name: "invalid module",
+			spec: &opa.OpaSpec{
+				Rego:    "package validate\n\nvalidate = true if true",
+				Modules: map[string]string{"empty": "testdata/empty.rego"},
```
brandtkeller, line 39: Should this be:
```suggestion
				Modules: map[string]string{"empty": "testdata/invalid.rego"},
```

src/pkg/providers/kyverno/types.go
brandtkeller: I believe that is a fair modification. Appreciate surfacing this change with context.

src/pkg/providers/opa/opa_test.go
nacx: Totally.

src/pkg/common/schemas/validation.json
```diff
@@ -523,6 +523,9 @@
                     "type": "string",
                     "pattern": ".*\\S\\s\\n.*"
                 },
+                "modules": {
+                    "type": "object",
```
brandtkeller: I'm getting an error for schema being valid with a quick [schema validator](https://gist.github.com/brandtkeller/4b998b2630235b8c3787da9a4a44fb0f). 

We'll evaluate getting this built into Lula - but error:
> Schema is invalid: failed to parse schema: invalid character '}' looking for beginning of object key string 

```suggestion
                    "type": "object"
```

brandtkeller (changes requested):
very minor schema adjustment (I'll file an issue to integrate schema validation more natively) otherwise this looks good.

src/pkg/common/schemas/validation.json
nacx: Thx!

brandtkeller (dismissed):
Looks great. Appreciate the minor adjustments. Will get an additional set of eyes on this so that we can merge.

meganwolf0:
Love this! Shout out to the [first PR](https://github.com/defenseunicorns/lula/pull/263) I put up for Lula that was attempting to do this (your implementation is much nicer!)

Pulled this down, tested, all lgtm, the only thing I might call out in the docs (or could probably be added in the code) is that `validate.rego` should be a protected key, I think it will override the policy if someone happens to specify a module key with the same text.

nacx:
Thanks!
You're right. I can change the code to add the user modules first, then the validate.rego to make sure it is not messed up. WDYT?
I'm out until 11th, though, I'll do the changes once I'm back.

src/pkg/providers/opa/types.go
```diff
@@ -59,6 +106,9 @@ func (o OpaProvider) Evaluate(resources types.DomainResources) (types.Result, er
 type OpaSpec struct {
 	// Required: Rego is the OPA policy
 	Rego string `json:"rego" yaml:"rego"`
+	// Optional: Modules is a map of additional OPA modules to include. The key is the name of the
```
zirain, line 115: should we point that `do not use validate.rego as module key`?

nacx: I'll add the comment, but also change how we put the modules in the map to make sure the validate rego does not get overwritten by user modules

meganwolf0:
> You're right. I can change the code to add the user modules first, then the validate.rego to make sure it is not messed up.

Sounds great, plus the doc point that zirain pointed out should cover that edge case. Thank you!

Had another thought, and this certainly shouldn't block the current PR but more of a future point of expansion perhaps - if you are executing many validations that are all using the same common module it'll be repeatedly downloaded; might be nice to look at caching that data across validations somehow. Might add as a follow-on issue once this is merged 😄

brandtkeller:
> > You're right. I can change the code to add the user modules first, then the validate.rego to make sure it is not messed up.
> 
> Sounds great, plus the doc point that zirain pointed out should cover that edge case. Thank you!
> 
> Had another thought, and this certainly shouldn't block the current PR but more of a future point of expansion perhaps - if you are executing many validations that are all using the same common module it'll be repeatedly downloaded; might be nice to look at caching that data across validations somehow. Might add as a follow-on issue once this is merged 😄

I'd agree - I think we can look at accomplishing this from multiple levels. One is general file caching within Lula - or more appropriately how do we "package" everything for portability. The other is the module references which are not retained across validations. Feels like a few issues worth scoping separately.

nacx:
Hi! I'm back and have addressed the review comments :)
I've opted to return a clear error if the `validate.rego` is used as a custom module name, to make sure validations don't run with an unexpected configuration leading to inaccurate results.

I agree caching would be great, but I'd like to keep that out of the scope of this PR.

brandtkeller (approved):
Great catch on the reserved module name @meganwolf0

meganwolf0 (approved):

Issue #730, opened 2024-10-11
Resolve code scanning findings in our workflows


brandtkeller:
### Describe what should be investigated or refactored

https://github.com/defenseunicorns/lula/security/code-scanning

### Expected Deliverable
Address the open scanning issues:

- [x] token-permission findings (fixed in https://github.com/defenseunicorns/lula/pull/830)
- [ ] branch protection error
- [x] SAST (dismissed: error was for old commits, SAST runs on all commits)  
- [x] #835
- [x] #836

mildwonkey:
Opened a PR for the easily-fixable issues - it was a lot of complaints about a single file missing top-level permissions.

Suggestions for the remaining issues: 

I think the SAST issue can be dismissed - it's complaining about a couple of commits that didn't have scans run, one being the first commit and another being a docs-only commit. (SAST runs on every PR now, so it shouldn't happen again).

The Fuzzing and CII Badge vulns should be converted to issues and worked on - are there any objections to working through  getting the badge? https://www.bestpractices.dev/en

Branch protections - ??? also an issue for someone to look into; I don't have access to the branch permissions so will need assistance; do we have a standard set of branch protections we're happy with for defense unicorns?

mildwonkey:
@brandtkeller if those ^^ look good to you I'll create issues, attach them to this one, and start working through! Just wanted feedback on the plan before I commit to too much work or sign up for something I shouldn't (sign up == log into https://www.bestpractices.dev/en and point it at lula)

brandtkeller:
Those sound great - let's get issues tracked so that we understand amount of work required.

mildwonkey:
The CII best practices is what I'm assuming will be the biggest one, because first we have to jump through the "access to our organization" hoops - I wonder if Zarf had to wait until it was out of the defense unicorns org to start that process 🤔 

The fuzzing remediate is done* pending approval and I'll ask our github admins about the current branch protection rules  



*done like all testing is "done": it's in there, but now it's on us to remember that it exists and add fuzzing tests when we write methods that are good candidates.

mildwonkey:
🤦🏻 I finally noticed that the specific branch protection issues are called out in the issue; I suspect we are ok with these settings for now (our gh administration is well locked down so I'm not worried about the first point, and 1 approving review feels good enough for the time being, though I certainly won't argue if we change it): 

```
* Warn: 'branch protection settings apply to administrators' is disable on branch 'main'
[...]

PR #832, opened 2024-12-03
chore: add fuzz tests and daily fuzz workflow

### LINKED ISSUE #835: Fuzzing

mildwonkey:
## Description
Remediating the "fuzzing" vulnerability in the security findings 

**NOTE FOR REVIEWERS** - I also turned off the errcheck linter for tests:
> I got _really annoyed_ at the errcheck linter complaining about tests not checking for errors: all of the fuzz tests are expected to MAYBE produce an error, and tbh I would've turned this linter off for tests sooner if I'd known how. But good news, now I know how to turn off specific linters just for `_test.go` files. If we don't like that, I'll remove the change and add the exceptions to each test, but again I regret not doing this from the start 😂  

I added some fuzz tests, using the very precise (sorry, "precise") heuristic of eyeballing a bunch of files and going with my gut. If a method dealt with bytes or arrays I gave it a closer look. (I also wrote some tests and deleted them later because they were very low value: not every byte is a good candidate for fuzzing). This is less of a "mission complete" PR and more of a "groundwork and examples for future-us" PR. 

Fuzzing tests can be slow and (at least for now) annoying to run: they can't be parallelized, and we can't even run multiple fuzz tests in a single package in a single command - they need to be explicitly run, one at a time, or you get errors like this and need to add `$` to the end of the test name to disambiguate:
 
`testing: will not fuzz, -fuzz matches more than one fuzz test: [FuzzCompareControls FuzzCompareControlsInt]"` 

Based on both of those, I added the fuzzing tests as a daily cron job instead of running them on every PR. I made the length configurable but we should default it to whatever makes most sense in CI. I left it at 10s, which was the maximum I'd tolerate while running from my laptop, but maybe 30-60s is better for the daily? TBH I don't think fuzzing is hugely valuable for this project, because we work with a lot of concrete types and strings (instead of doing a lot of things with really random user input) and also write good tests in general, which is why I wanted this to be effective and useful, but not too intrusive. 

## Related Issue
Relates to #730
Closes #835 

## Type of change

### DIFF
--- a/.github/workflows/test-fuzz.yaml
+++ b/.github/workflows/test-fuzz.yaml
@@ -1,11 +1,11 @@
-name: Unit Testing
+name: Fuzz Testing
 
 permissions:
   contents: read
 
 on:
   schedule:
-    cron: 0 0 * * * 
+   - cron: '0 0 * * *' 
 
 jobs:
   test:
--- a/.github/workflows/test-fuzz.yaml
+++ b/.github/workflows/test-fuzz.yaml
@@ -1,0 +1,25 @@
+name: Unit Testing
+
+permissions:
+  contents: read
+
+on:
+  schedule:
+    cron: 0 0 * * * 
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+    - name: Setup golang
+      uses: ./.github/actions/golang
+
+    - name: Display Go version
+      run: go version
+      
+    - name: Fuzz Tests
+      run: | 
+        make test-fuzz
--- a/src/cmd/dev/validate_test.go
+++ b/src/cmd/dev/validate_test.go
@@ -1,0 +1,35 @@
+package dev
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func FuzzRunSingleValidation(f *testing.F) {
+	for _, tc := range []string{"../../test/unit/common/oscal/valid-component.yaml", "../../test/unit/common/oscal/valid-component-metadata-injected.yaml"} {
+		bytes, err := os.ReadFile(tc)
+		require.NoError(f, err)
+
+		f.Add(bytes)
+	}
+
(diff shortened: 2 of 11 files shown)


brandtkeller:
Going to invest some more time in this shortly - but overall appreciate the instrumenting of fuzzing

meganwolf0:
I like this! And this is just a general comment about testing, but I think it would be nice to have like a "testing best practices" or something doc, probably in community/contrib that talks briefly about when we expect certain kinds of tests for things (fuzzy ones, etc). IDK if this is a good PR to at least introduce the concept of that for fuzz tests in particular, but yeah just throwing it out there.

src/cmd/dev/validate_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func FuzzRunSingleValidation(f *testing.F) {
+	for _, tc := range []string{"../../test/unit/common/oscal/valid-component.yaml", "../../test/unit/common/oscal/valid-component-metadata-injected.yaml"} {
```
meganwolf0: Believe the inputs here should be down in `src/test/unit/common/validation/<pick-your-favorite>`

meganwolf0 (commented):
All lgtm! Only thing is a suggestion to change the fuzz sample in cmd/dev tests to a shape more like the validation.yaml, since that is the expected input there

.github/workflows/test-fuzz.yaml
```diff
@@ -0,0 +1,25 @@
+name: Unit Testing
+
+permissions:
+  contents: read
+
+on:
+  schedule:
+    cron: 0 0 * * *
```
meganwolf0: should this be a list? https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

mildwonkey: yes it should!

mildwonkey:
Hopefully that also fixed the job name not showing up in the workflows tab (it was using the file name)

meganwolf0 (approved):

brandtkeller (approved):
Appreciative of this being added! Looks like a great scope for first iteration and don't disagree with your thoughts.

PR #845, opened 2024-12-11
feat(validation-tests): validation tests path resolution

### LINKED ISSUE #840: Validation tests path not resolving as expected
### Steps to reproduce
1. Using a validation test, such as:
```yaml
tests:
  - test-update-seq
    expected-result: not-satisfied
    changes:
      - path: data.seq.[0].val
        type: update
        value: new-value
```
where the structure may be as such:
```json
{
  "data": {
      "seq": [
       { 
         "value": "old-value"
       }
     ]
  }
}
```

### Expected result
The test data for `test-update-seq` should identify the first item in `data.seq` (as given by [0]) and modify `value` to be `new-value`

### Actual Result
Item [0] is unable to be identified in `seq` because it assumes based on the path separation that `seq` is a map instead of a sequence type.

### Severity/Priority
High
Rationale - this is a new feature that's probably not adopted much _yet_, however getting the path semantics and fixes out ASAP would be good to guard against future issues/blockers to users implementing this functionality

### Additional Context

The most immediate fix for this is to identify anything numerical and apply a different filter in `internal/transforms`. However the downside of this approach is that valid map keys can be numerical, so if you are actually trying to access a map key instead of an index in a sequence that solution would not work.

The proposed solution is an update to the path parsing methods. So to access any map keys that are inside [] (which are needed for keys that include ".", for instance) the new assumptions/rules should be that when a [] is included, the preceeding path is assumed to be a map if the content inside the [] are incapsulated by double-quotes. Take the following examples:

* path.value[0] (== path.value.[0]) -> Identifies the first element in the `value` key, which is a list under the `path` map
* path.value["my-value"]  (== path.value.["my-value"] == path.value.my-value) -> Identifies the `my-value` key that exists in the `value`, which is a map type under the `path` map
* path.value.0 (== path.value["0"] == path.value.["0"]) -> Identifies the `0` key that exists in the `value`, which is a map type under the `path` map

> [!NOTE]
> Additionally, the "." between a key and right square bracket should be optional - this is an artifact of how kyaml path splits, but it's actually a very easy function to just replace. After working with the functionality a bit, it's much less semantic to add a "." between bracketed elements. This can also easily be made backward compatible so that both versions work

meganwolf0:
## Description

Updates the validation testing and fixes a bug with the way indicies were identified (see #840 ) for specific bug.
This PR includes 
* Re-factor of the existing filter/path functionality. Intent was to increase test coverage and generally make this package easier to understand/use
* Additional functionality to delete an item from a list (this was not previously supported). Solving the bug above yielded some code that made this functionality pretty easy to implement.
* Updated the path parsing to make it more semantic for users - previously all items needed to be "." delimited, now made it so `key[foo=bar]` is equivalent to `key.[foo=bar]`
* DOCS! Added some new docs for testing a validation, as I felt that was missing context. Also updated the testing.md reference doc.

## Related Issue

Fixes #840

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging

- [ ] (N/A)[Schema Updates](https://github.com/defenseunicorns/lula/blob/main/docs/community-and-contribution/schema-updates.md) applied


docs/getting-started/develop-a-validation.md
```diff
@@ -247,6 +248,8 @@ Here, we will step through creating a sample validation using the Kubernetes dom
         ```
 
         Success! Additional conditions can be tested this way to fully stress-test the validity of the policy.
+    
+    3) **Preferred Approach** Implement tests natively in the validation, using the [testing guide](../reference/testing.md). See the next tuturial [test a validation](./test-a-validation.md) for more information.
```
mildwonkey: typo!
```suggestion
    3) **Preferred Approach** Implement tests natively in the validation, using the [testing guide](../reference/testing.md). See the next tutorial [test a validation](./test-a-validation.md) for more information.
```

docs/getting-started/test-a-validation.md
```diff
@@ -0,0 +1,263 @@
...
+  value-map:
+    replicas: 0
+```
+
+This is another test case where at first pass we'd expect to set up a change as follows:
+
+```yaml
+- path: podinfoDeployment.status.replicas
+  type: update
+  value: "0"
```
mildwonkey, line 142: I might add a comment inline with this example to make it very clear that this version won't work - it's easy to start copy pasting snippets out of tutorials without realizing you're copying the "don't do it this way" version if it isn't labeled incredibly clearly

docs/getting-started/test-a-validation.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+However, this will not correctly generate the expected json structure, as the `replicas` field is a number, and not a string. Instead, we need to use the `value-map` change type, which allows us to set the value of a field to any type of value.
```
mildwonkey: If you're doing to do show a "not working" snippet, I think you should add also show the working snippet before putting it all together in the full example below - if you look at just the snippets and not the words we're telling a confusing story. (this isn't quite formatted properly as a suggestion)

```suggestion
However, this will not correctly generate the expected json structure, as the `replicas` field is a number, and not a string. Instead, we need to use the `value-map` change type, which allows us to set the value of a field to any type of value:
```

Then add the correct snippet below (formatting got weird, you get the idea)

```yaml
- path: podinfoDeployment.status
   type: update
   value-map: 
      replicas: 0
```

docs/reference/testing.md
```diff
@@ -50,69 +52,101 @@ tests:
...
 
 >[!IMPORTANT]
 > The path will return only one item, the first item that matches the filters along the path. If no items match the filters, the path will return an empty map.
 
+#### Path Rules
+* Path resolution supports both `path.[key=value]` and `path[key=value]` syntax
+* In addition to simple selectors for a list, e.g., `path[key=value]`, complex filters can be used, e.g., `path[key=value,key2=value2]` or `path[key.subkey=value]`
+* Use double quotes to access keys that contain periods, e.g., `foo["some.key"=value]` or `foo["some.key/label"]`
+* To access the index of a list, use `[0]` (where 0 is any valid index) or `[-]` for the last item in the list
+* If you need to access a map key that is an integer, either use `foo["0"]` or `foo.0`
```
mildwonkey: it's not that I don't believe you, but the `"0" in ``foo["0"]` does not look like an integer  - this is confusing! To access a map key that is an integer, pass in a string????? 😭

mildwonkey (changes requested):
One typo to fix, the other suggestions you are free to take or ignore as you see fit! This looks good, and I like your documentation.

docs/getting-started/test-a-validation.md
meganwolf0: I showed the "right" way above, are you just thinking from like a better story-telling perspective it makes sense to start with the "wrong" way - like "a user may be inclined to do this thing <wrong way>, DONT do this thing, instead do this thing <right way>"?

docs/reference/testing.md
meganwolf0: sorry, that is confusing wording. I guess I should say a string that only contains an integer? Because I don't think integer types are actually valid keys haha, so that's what I was after. Idk if you have some better words for that

mildwonkey: Yeah, map keys are required to be strings, so it's a confusing statement. Might be clearer to just say "map keys must be strings so if you have an integer, wrap it in quotes"? or just "map keys are strings"

docs/getting-started/test-a-validation.md
mildwonkey: 🤔 My mistake, sorry! I clearly misread the document, I thought the "right way" you mentioned above was still part of the previous example. I'll resolve this 'cuz I don't have a suggestion, but I find it confusing (as someone to whom this is all new syntax) to be told how I would expect to do something, when that's not what I was thinking when reading that section. 

I'd personally like to see a comment in the invalid snippet that says `# invalid syntax` or something but now I'm just exposing that I look at the code snippets without reading 😂

mildwonkey (dismissed):
LGTM 👍🏻 

(I'll re-approve if you want to change the integer map key wording but my actual requested changes are resolved)

mildwonkey (approved):
I like the changes! 😀 ❤️ :shipit:

brandtkeller (approved):
I really like the `test-a-validation.md` doc.

Issue #776, opened 2024-11-02
Improve traceability of controls with lack of validation/evidence


brandtkeller:
### Is your feature request related to a problem? Please describe.

By default - Lula takes a stance on being accessible to existing OSCAL such that - on validate - any `implemented-requirements` without a lula validation (and therefor any proof of compliance) will result in a `not-satisfied` finding without any `related-observations` to indicate no observations performed. 

When you combine 1 -> N components with 1 -> N control-implementations with 1 -> N potentially overlapping controls - this results in some confusion. 

For a single finding - all related observations may be in a `satisfied` state - yet be failing because the control is specified in another component and has no validations and therefor is turning the finding to `not-satisfied` without any related observation to indicate why. 

### Describe alternatives you've considered

Change the default stance

### Additional context

For `implemented-requirements` without any Lula validations - it may be worth experimenting with adding an observation to indicate 

### Expected Deliverable

- [ ] Support for better traceability across the assessment result finding state.

meganwolf0:
Feels similar to #350

meganwolf0:
I think #847 and #862 set up the framework to account for this issue, but I believe an additional appropriate `reason` that would cover down on what this specific issue is referring to might be something like "some component linked to this control is lacking a Lula Validation" - since this case is where all validations are passing, but there's a gap such that all components do not have validations. This might take some additional work to instrument, as I'm not sure the fidelity of the information which would provide this rationale is currently tracked.

If this is indeed the agreed fix for this issue, it might be best to close this and reopen a more finely scoped issue for that specific improvement.

brandtkeller:
I am unclear if we need another `reason` or if we need to build more context with current `remarks`. 

IE we have reasons `pass`, `fail`, and `other` with some remarks indicating context for the decision. 

There is still a gap with the remarks based on the order of multiple components with the same control that we should address. Open to thoughts on which field is modified. I'd wager we could update `remarks` and close this out.

## Gap

Based on the logic in #862 - there is still a gap where a finding would have `related-observations` with a status of:
```
status:
  reason: other
  remarks: No Lula validations were defined for this control
  state: not-satisfied
```

This is likely due to the sequence of processing multiple components with the same control and first processing a finding without any validations - then processing the same `control-id` with validations.

meganwolf0:
yep sorry, `remarks` is right -

[...]

Issue #350, opened 2024-04-01
Assessment-Results Finding Reason


brandtkeller:
Support for adding `reason` and `remarks` to the `assessment-results.results[_].findings[_].target.status`

## Objective
`state` is already implemented. `reason` allows for `pass`/`fail`/`other`. If possible we should map information back to remarks to indicate how the `state`/`reason` came to be. 

Additional context - Lula to-date provides a finding for all `implemented-requirements` - In the event there is no lula validation - this would be a great thread for a default behavior with reason: "other".

meganwolf0:
Believe this is closed by #847 - @brandtkeller for confirmation

PR #847, opened 2024-12-12
feat(valiation): account for controls not evaluated by Lula

nacx:
## Description

Opening as a draft to discuss the more convenient way to provide feedback when some controls do not have any Lula validation defined:

Currently, if the implemented controls do not define a Lula validation, they are reported as `not-satisfied`. This is fine, but it would be great to indicate the reasons for that verdict in the assessment results, such as "no lula validation is defined for this control", so users can properly reason about the result.

This can be achieved in different ways I'd like to discuss:

* In the first place, is this a valid use case? Should the project account for such implemented-controls that do not define a lula validation? (And if not, should there be some validation instead?)
* If the answer to the above is that Lula should gracefully handle that:
  * Does it make sense to keep findings for those controls in the findings list and in the reviewed-controls?
    * If not, the current state of the PR addresses it by not including the findings in the result. Or should that be reported in a different way?
    * If yes, what would be the best way to communicate that no validation was performed to determine the control is not-satisfied? Would a note in the finding remarks be OK? Or should that information be provided somewhere else?

## Related Issue

https://github.com/defenseunicorns/lula/issues/776
https://github.com/defenseunicorns/lula/issues/350

## Type of change

## Checklist before merging


brandtkeller:
> Currently, if the implemented controls do not define a Lula validation, they are reported as `not-satisfied`. This is fine, but it would be great to indicate the reasons for that verdict in the assessment results, such as "no lula validation is defined for this control", so users can properly reason about the result.

Agreed!

> This can be achieved in different ways I'd like to discuss:
> 
> * In the first place, is this a valid use case? Should the project account for such implemented-controls that do not define a lula validation? (And if not, should there be some validation instead?)

I believe the project should still account for them - more context below. 

> * If the answer to the above is that Lula should gracefully handle that:
>   
>   * Does it make sense to keep findings for those controls in the findings list and in the reviewed-controls?

Current line of thought is yes - Retaining these controls as `Findings` allows for a few core principles to Lula adoption:
1. Lula augments OSCAL - People can bring their own OSCAL with zero Lula implementation and generate an Assessment Results file that they could then use manually or with other systems. 
2. It allows for historical traceability - A control that was `not-satisfied` previously is now `satisfied` -> Report this (as well as the opposite scenario)
     
>     * If not, the current state of the PR addresses it by not including the findings in the result. Or should that be reported in a different way?
>     * If yes, what would be the best way to communicate that no validation was performed to determine the control is not-satisfied? Would a note in the finding remarks be OK? Or should that information be provided somewhere else?

I believe `finding.remarks` we be a great place to quickly annotated that no evidence was provided or collected to support the control being satisfied. This allows us to explore more concepts around what evidence means to findings. 

Open to other opinions for consideration though.

nacx:
>I believe finding.remarks we be a great place to quickly annotated that no evidence was provided or collected to support the control being satisfied. This allows us to explore more concepts around what evidence means to findings.

That sounds good to me and as a good starting point. I'll update the PR accordingly following this approach and add the relevant tests, etc. Thx!

meganwolf0:
I'd possibly throw out the other field in `status` which is the `reason` one. I sort of interpreted that field as more "machine readable", with remarks being just additional context. The constraint is that the value "may be locally defined", and I'm not totally sure what that is, but I assume it's something we could choose to enumerate however we want?

So for example we could have an enumeration of options for reason (validations-failed, validations-passed, partial-validations, no-validations) then some additional more human friendly remark that's like "No validations are linked to the control" or "Some validations failed (this one, that one)"

This is also something we could instrument over time and for this initial thought I think remarks would suffice 😄

brandtkeller:
Ah great point - I think it would be great to start the implementation of `reason`. The `token` type and ambiguity around "may be locally defined" (I need to understand what this really means) makes this a little unclear. 

That said - leaning into `finding.status.reason` = `other` as defined natively and `finding.status.remarks` being the remarks section to augment?

meganwolf0:
> That said - leaning into finding.status.reason = other as defined natively and finding.status.remarks being the remarks section to augment?

Yeah I think that would work well!

nacx:
Ok, pushed a commit implementing the suggested approach

src/pkg/common/requirement-store/requirement-store.go
```diff
@@ -96,19 +96,31 @@ func (r *RequirementStore) GenerateFindings(validationStore *validationstore.Val
...
+		} else if pass > 0 && fail == 0 {
 			state = "satisfied"
+			reason = "pass"
+		} else if pass == 0 && fail == 0 {
+			// If there is no result (pass or fail) it means that no validation was performed by Lula.
+			// When that happens we can explicitly add a note to the finding, to properly explain the
+			// reason for the control being not-satisfied
+			state = "not-satisfied"
+			reason = "other"
+			remarks = "No Lula validations were defined for this control"
```
brandtkeller, line 112: @meganwolf0 @mildwonkey How do you feel about this language vs "No evidence was provided for this control". 

Proposed language is %100 correct - but curious on the stance that Lula augments OSCAL as opposed to controls it inherently - if we might take a stance on Evidence being the driving factor - emphasizing what Lula validations provide. 

Would this confuse a user potentially?

meganwolf0: hmm I'm of two minds on this one - agreed that "No evidence was provided..." is more flexible/lula agnostic, which I like. On the other hand, Lula _is_ actually determining that the control is `not-satisfied` because there are no lula validations, we don't really support another concept of or evaluation of other "evidence" generically. So I guess in the interest of being more clear to a user*, I might defer to actually stating why something was marked as `not-satisfied`, which I think the current statement does more accurately. 

*I suppose in terms of "confusing the user", I think if the user is the one executing Lula, the current text is less confusing. If the user is some downstream consumer of the product, then yeah maybe that text is more confusing if they don't understand what lula is/provides 🤔 tough one haha

mildwonkey: I agree with @meganwolf0 - if/when we refactor Lula to read attestations/validations/evidence from other sources can refactor the error messages, but right now this seems to be the clearest and most accurate.

brandtkeller: Resolving as evidence ( 😉 ) for how we can use more agnostic language that the industry/auditors will more closely understand in the future.

brandtkeller (approved):

mildwonkey (approved):
👍🏻 LGTM, thanks again for the submission! Keep 'em coming!!! 😂 ❤️

PR #851, opened 2024-12-12
API chaining POC

### LINKED ISSUE #841: API Domain: chaining requests
## Description

The API domain’s current functionality supports basic Get and Post API operations. These are insufficient to support common API operations. Even basic API operations may require multiple API calls: some examples include logging in to get an API token, listing all resources before acting on individual resources, and getting IDs from resource names. And even if the API plays nicely with Lula’s current functionality, many validations require more than a single call, and users frequently may need to mutate one API call output for another's input.

## Additional Context

While we can solve the API Chaining request today, limitations in Lula’s architecture and configuration language are increasingly holding us back from the most performant and effective solutions. This is therefore a two pronged proposal: implement an experimental API Chaining feature, then tackle the larger  Configuration language/DSL refactor so that we can continue to build on the API chaining (and everything else which will benefit). The actual Configuration Language refactor will be a separate (future) proposal.

1. Implement API Request chaining using the simplest and quickest implementation (that still meets our needs and quality/security requirements)
2. Follow this up by refactoring Lula to use a fully featured configuration language & DSL which will address the various limitations of the selected API chaining solution.

mildwonkey:
POC: "templatizing" URL, headers, and parameters in the API domain with results from previous API calls.

### DIFF
--- a/src/pkg/domains/api/api.go
+++ b/src/pkg/domains/api/api.go
@@ -49,7 +49,12 @@ func (a ApiDomain) makeRequests(ctx context.Context) (types.DomainResources, err
 		for i, request := range a.Spec.Requests {
 			if i > 0 { // the first request cannot use outputs
 				if a.Spec.outputs != nil {
-					executeTpls(request, a.Spec.outputs)
+					var err error
+					request, err = executeTpls(request, a.Spec.outputs)
+					if err != nil {
+						/// TODO: better error
+						return collection, err
+					}
 				}
 			}
 
@@ -82,7 +87,6 @@ func (a ApiDomain) makeRequests(ctx context.Context) (types.DomainResources, err
 				}
 
 				if request.Outputs != nil {
-					// get the value from the dr, it's already closer to yaml-shaped
 					node, err := yaml.ConvertJSONToYamlNode(string(response.Raw))
 					if err != nil {
 						errs = errors.Join(errs, err)
@@ -117,6 +121,7 @@ func executeTpls(req Request, vars map[string]map[string]interface{}) (Request, 
 		reqUrl, err := url.Parse(urlstr)
 		if err != nil {
 			// TODO: error types! this is a copy!
+			// TODO: we should attempt each template step and return a wrapped error
 			return modifiedReq, errors.New("invalid request url")
 		}
 		modifiedReq.reqURL = reqUrl
--- a/src/pkg/domains/api/api.go
+++ b/src/pkg/domains/api/api.go
@@ -143,7 +143,7 @@ func executeTpls(req Request, vars map[string]map[string]interface{}) (Request, 
 	// params
 	if req.ParamsTpl != nil {
 		params := make(map[string]string)
-		for k, v := range req.Options.HeadersTpl {
+		for k, v := range req.ParamsTpl {
 			h, err := executeTpl(v, vars)
 			if err != nil {
 				return modifiedReq, err
--- a/src/pkg/domains/api/api.go
+++ b/src/pkg/domains/api/api.go
@@ -13,6 +13,7 @@ import (
 
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 
+	"github.com/defenseunicorns/lula/src/pkg/message"
 	"github.com/defenseunicorns/lula/src/types"
 )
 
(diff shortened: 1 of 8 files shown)


src/pkg/domains/api/api.go
```diff
@@ -41,7 +45,18 @@ func (a ApiDomain) makeRequests(ctx context.Context) (types.DomainResources, err
 		// requests with overrides (in request.Options.Headers) will get bespoke clients.
 		defaultClient := clientFromOpts(defaultOpts)
 		var errs error
-		for _, request := range a.Spec.Requests {
+		for i, request := range a.Spec.Requests {
+			if i > 0 { // the first request cannot use outputs
```
soltysh, line 51: Nit: but I've noticed this being heavily used in lula codebased, but something that makes me :crying_cat_face: 
how about rewriting this like so:
```golang
for i, request := range a.Spec.Requests {
    if i < 1 {
        // the first request cannot use outputs
        continue
    }
    if a.Spec.outputs != nil {
	    var err error
...
```
The benefit is that we're having less indented code which improves readability.

src/pkg/domains/api/api.go
```diff
@@ -41,7 +45,18 @@ func (a ApiDomain) makeRequests(ctx context.Context) (types.DomainResources, err
...
+				if a.Spec.outputs != nil {
+					var err error
+					request, err = executeTpls(request, a.Spec.outputs)
+					if err != nil {
+						/// TODO: better error
+						return collection, err
```
soltysh, line 57: Consider if you want to process all the request and gather `[]error` and return that, or fail on the first error like you have here.

src/pkg/domains/api/types.go
```diff
@@ -42,8 +43,13 @@ type ApiSpec struct {
 	// specified ApiOpts
 	Options *ApiOpts `mapstructure:"options" json:"options,omitempty" yaml:"options,omitempty"`
 
-	// internally-managed fields executable will be set to true during spec
-	// validation if *any* of the requests are flagged executable
+	// internally-managed fields
+	//
+	// outputs is a map of stored outputs from API calls that can be used as
+	// template variables for URLTpls
+	outputs map[string]map[string]interface{}
```
soltysh, line 50: Several ideas, take whatever you feel like makes sense:
1. I'd not introduce this intermediary in the external types at all. Even though it's a private field, it still feels like exposing part of internal code. I'd suggest creating an internal structure, if you need to pass the data between one data structure and the other. 
2. I looked into the templating bits and I have similar concerns as you pointed out on slack, that in the long run this might cause pains. I might be missing some bits of information about the goal, but have you considered using jsonpath, which is already used in several places and can potentially provide a better escaping logic, and be easier to understand by users?

src/pkg/domains/api/api.go
mildwonkey: That's the current design, and there's more (not yet implemented) nuance to the statement I'm about to make, but once we introduce templating we need to fail as soon as one API call fails, since the output of that call is the input of another. We cannot complete a chain of requests if a request in that chain fails.

I will say (here's the nuance) that I'd like refactor Lula so it has an idea of a request chain vs a single request, or some other way to identify while operations much happen first and succeed before the next can continue. This POC is meant to be the API chaining option that doesn't require a huge refactor, but I'm team refactor, myself.

src/pkg/domains/api/types.go
mildwonkey: That is basically the syntax I have for users to define what fields they want from the previous API calls (i'm using the kyaml library, for consistency with other parts of lula). 

That doesn't help me with the user input template part of things, though - since we're "templatizing" strings (parameters, headers, the URL) I do need some kind of string-based templating syntax?

src/pkg/domains/api/api.go
mildwonkey: This is also on my list of arguments for a DSL : the go-cty type system has Unknown and Dynamic `type` placeholders, and I want that or something like it in Lula, so it's easier to look at a single API request and know that we're still waiting for placeholder values to be expanded. (we'd also need a dependency graph)(i know i'm just describing terraform, i'm sorry)

src/pkg/domains/api/api.go
mildwonkey: I'm onboard with the idea of reducing indentation, but that won't work in this example - here I'm not continuing out of the entire loop, just skipping the outputs processing (those first 10 lines).

brandtkeller:
Linked current work (here) with associated issue for future reference.

PR #853, opened 2024-12-12
fix(oscal): update version of assessment results generated

brandtkeller:
## Description

Noticed the `assessment-results` object output from `validate` was not version `1.1.3` due to a constant we have defined. Bumping this version to match intended version. 

## Related Issue

No issue - noticed a part of the workflow missing update from the transition to OSCAL `1.1.3` artifacts. 

## Type of change

## Checklist before merging


meganwolf0 (approved):
I see a `latestVersion` const in go-oscal - wonder if it might make sense to make that public instead and use that in the future?

brandtkeller:
> I see a `latestVersion` const in go-oscal - wonder if it might make sense to make that public instead and use that in the future?

Worth consideration - It would remove potential drift when updating - but also remove some inherent control that a consumer of the library has over updating the dependency with additional plan to work through some supporting behaviors before consuming the type updates.

PR #862, opened 2024-12-19
fix(assessment): adjust reason accordingly across multiple components

### LINKED ISSUE #861: Improve clarity around newly implemented "reason"
There is a minor gap in the implementation of `finding.status.reason` where currently we report `reason: other` for a finding that has no lula validations associated - but when there is multiple components where one implemented requirement with the same control-id has validations and the other does not - this is still being reported as `other` instead of failing. The remarks then say "No Lula validations were defined for this control" when there is indeed observations.

### Severity/Priority

Low

brandtkeller:
## Description

Minor gap found when testing across multiple components with the same `control-id` implemented-requirement.

## Related Issue

Fixes #861 

## Type of change

## Checklist before merging


CloudBeard (approved):
Great change, this should make it easier to find things in the assessment results manually and setup for report/console/etc later on too.

PR #869, opened 2025-01-08
chore: add debug-level logging to doHTTPReq

mildwonkey:
adding the logging statements I need in all my current POC work so I don't have to keep carrying commits between branches 😂


brandtkeller (approved):
Appreciate the initiative - given logging discussions I think this is reasonable.

PR #874, opened 2025-01-13
feat(oscal)!: oscalmodel interface implementation for component definitions

### LINKED ISSUE #747: OSCALModel interface implementation for Component Definitions

As proposed in #694 - support updating component definition use to include the interface implementation 

### Expected Deliverable

- [ ] Component Definition type with required methods implemented

meganwolf0:
## Description

Implementation of the OSCALModel interface for a Component Definition. Additionally:
* Updated implementation of component definition in `generate`
* Added types constants for the models in `complete-schema`
* Updated the ssp `HandleExisting` function as an error was noted with variable overwriting (+ added test)

**Note** Skipping function updates to compose/validate to use this interface. As those are probably wrapped more directly in the validation refactor, opting to defer updates to those issues.

## Related Issue

Fixes #747

## Type of change

## Checklist before merging


src/pkg/common/oscal/component_test.go
```diff
@@ -112,13 +115,14 @@ func TestNewOscalComponentDefinition(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := oscal.NewOscalComponentDefinition(tt.data)
+			model := oscal.NewComponentDefinition()
+			err := model.NewModel(tt.data)
```
mildwonkey, line 119: massively nitpicky, feel free to ignore, but I'd rename this TestNewModel, since NewOscalComponentDefinition is now just returning an empty struct and the test is actually exercising NewModel

src/pkg/common/oscal/system-security-plan.go
```diff
@@ -81,12 +81,12 @@ func (ssp *SystemSecurityPlan) HandleExisting(path string) error {
 		if err != nil {
 			return fmt.Errorf("error reading file: %v", err)
 		}
-		ssp := NewSystemSecurityPlan()
-		err = ssp.NewModel(existingFileBytes)
+		newSsp := NewSystemSecurityPlan()
```
mildwonkey, line 84: This feels a bit awkward; do we most commonly call NewThing followed by NewModel? If so, can this be refactored so that it can be written in one line? 

Some ideas -  

`model, err := NewSystemSecurityPlan().NewModel(existingFileBytes)`
If the above doesn't work out (it might not make sense, given the interface) a single method, NewSystemSecurityPlanWithModel, could also make for a clearer developer experience

mildwonkey (dismissed):
I'm sorry, I thought I submitted this yesterday!  

I left some non-blocking nitpicks - you're free to ignore them (or address in a later PR, I know you have a few queued up)

brandtkeller (dismissed):

src/pkg/common/oscal/system-security-plan.go
meganwolf0: yeah NewModel doesn't return the ssp at the moment, so that signature would have to be modified. I suppose I think the reason for the break out is such that if you're generating an ssp, you don't have the existing data (I think a maybe clearer name for the `NewModel` method would be like `NewModelFromBytes` or something to indicate maybe there's multiple avenues to obtaining a new model)

meganwolf0: on the second part, which I glazed over in the response, `NewSystemSecurityPlanWithModel`... hrm would that be generalized to the interface as well?

mildwonkey: 🤔 I would assume  `NewSystemSecurityPlanWithModel` would be a standalone function, not an interface method, but honestly whatever makes sense is fine (even if that's "change nothing for now")

mildwonkey (approved):

PR #876, opened 2025-01-14
feat(oscal)!: oscalmodel for assessment results

### LINKED ISSUE #748: OSCALModel interface implementation for Assessment Results

As proposed in #694 - support updating assessment results use to include the interface implementation 

### Expected Deliverable

- [ ] assessment results type with required methods implemented

meganwolf0:
## Description

Implementation of the OSCALModel interface for a Assessment Results

## Related Issue

Fixes #748 

## Type of change

## Checklist before merging


src/pkg/common/oscal/complete-schema.go
```diff
@@ -18,6 +18,16 @@ import (
 	"github.com/defenseunicorns/lula/src/pkg/message"
 )
 
+const (
+	OSCAL_COMPONENT            = "component"
```
mildwonkey, line 22: Can I ask why you're using all caps for these? There's nothing wrong with it, it's just not something I see in go a lot, so when I see it I tend to expect more meaning/functionality behind those particular consts. 

----
This is definitely me over engineering for the future, but you also may consider defining a type  - 

I don't know if this is useful for you today, so feel free to ignore it - I'm thinking about future design ideas where a type switch on oscal thing (I keep wanting to say OSCAL component, to refer to a block of OSCAL. WHAT'S THE RIGHT WORD HERE? Oscal ... noun?) would be helpful.

```
type OscalType string

const (
  Component OscalType "component"
 ... etc
)
``` 

If we do go in that direction, I'd use stringer and use an int for the type, though that's more habit than anything: https://pkg.go.dev/golang.org/x/tools/cmd/stringer

mildwonkey (approved):

src/pkg/common/oscal/complete-schema.go
mildwonkey: ^^ none of this is blocking, just a question

meganwolf0: On the caps thing... I don't know, I guess I did think all caps constants as a convention? We do use all caps constants elsewhere too (where I didn't add them haha). But that might be an artifact of other language devs using go, so if that's like an anti-pattern in go constants certainly happy to rename!

As for the other point about just using a stringer there... yeah I think you might be right that it would be a better implementation. Trying to think of an actual reason for hardcoded strings being "component", "assesssment-results", etc, and I don't actually see one. @brandtkeller can you think of a reason the types actually need to return those string values, I can't - other than "that's just what we had"

brandtkeller: No core requirements in particular - more of a mock for proving out the interface and showing some skeleton that may or may not have any relevance as we implement.

meganwolf0: alright back to this - I looked at the change and it's actually touching more in the codebase to swap all those types. I think this is something we should do after migrating all the models to the OscalModel interface, since at least a few of the methods that rely on those strings will probably cease to exist

Issue #810, opened 2024-11-20
Partial Component Compositions


meganwolf0:
### Is your feature request related to a problem? Please describe.

Refactor the component-definition composition functionality to only compose components but maintain validations and relative links.

Some of this work might relate to/help uncover the solutions for #716 and #717 

### Describe the solution you'd like

- **Given** component defn with imported components and validation links
- **When** composition is executed on said component defn with optionality to only compose the components
- **Then** validations are not incorporated in the back-matter; relative path links are handled

### Describe alternatives you've considered

Add all validations to the back-matter in component defn and then just add those to the back-matter in the SSP

### Additional context

To enable the SSP to use the validations, while they are still in a disaggregated state (i.e., not in the back-matter), the component-definition needs to be passed to the SSP generate processes with those links intact and properly remapped relative to the SSP output file path.

### Expected Deliverable

- [ ] Updated optionality on composer 
- [ ] Updated documentation/intent on `tools compose`, intended use cases, scenarios (coupled to #715). Where does `decompose` fit? (#397)

meganwolf0:
Closed via #905

PR #879, opened 2025-01-15
feat(oscal): link remapper method component defn

### LINKED ISSUE #878: Link re-mapper method on Component Definition

Investigate and implement a method that can remap relative references contained in a component definition. Intent to support composing component definitions, without importing the related validations into the backmatter. Additionally could have some utility when a `packaging` feature is implemented in the future.

- **Given** a component defn references several file paths
- **When** link re-mapping method is executed with a new directory location
- **Then** file paths are re-written such that they are relative to the new directory

### Additional context

Might warrant an update to the interface, TBD.

### Expected Deliverable

- [ ] New link re-mapper method on ComponentDefinition + Unit test

meganwolf0:
## Description

Added a new method to component definition - `RewritePaths`

Intent of this method is to facilitate aggregation of data in component definitions (e.g., via Compose functionality) where the references contained therein may be relative to said component definition. This method should also have extensibility to be used for a future `packaging` type workflow where we may wish to pull disaggregated OSCAL models into a centralized, opinionated structure. Hence, this method may make sense to add to the interface when it's been implemented across models.

**Notes to reviewers**
* I updated the network function `GetLocalFileDir` -> before it was doing a file existence check which I think is probably not needed in this function itself. Removal allowed for easier testing. Also needed to update this function to check for the opaque structure (before was just checking host) - see [net/url](https://pkg.go.dev/net/url#URL) for reference

## Related Issue

Fixes #878 

## Type of change

## Checklist before merging


src/pkg/common/common.go
```diff
@@ -211,3 +213,98 @@ func CleanMultilineString(str string) string {
...
+		// if val.Kind() is not a string, we can't update it
+		return fmt.Errorf("cannot update type %s", val.Kind())
+	}
+
+	return nil
+}
+
+func splitPath(path string) []string {
+	components := strings.Split(path, ".")
+	return components
```
mildwonkey: this can be a one-liner - probably not worth writing a method, but if you do, don't bother initializing the components variable, just return the output directly (`return strings.Split(path, "."`)

src/pkg/common/common.go
```diff
@@ -211,3 +213,98 @@ func CleanMultilineString(str string) string {
...
+// TraverseAndUpdatePaths uses reflection to traverse the obj based on the path and update file path references
+func TraverseAndUpdatePaths(obj interface{}, path string, baseDir string, newDir string) error {
+	// Split the path into components
+	components := splitPath(path)
+
+	// Start reflection traversal
+	return reflectTraverseAndUpdate(reflect.ValueOf(obj), components, baseDir, newDir)
+}
+
+func reflectTraverseAndUpdate(val reflect.Value, components []string, baseDir string, newDir string) error {
```
mildwonkey: This gives me a bit of anxiety - reflect is not a very performant package (and it moves more errors to runtime, away from compile time); it's something I generally avoid using in production code unless it's the only solutoin. That also doesn't mean it's not the correct choice, but it is something to be aware of. In the future we may want to add some benchmarks and keep an eye on how long it takes to process large documents.

src/pkg/common/common_test.go
```diff
@@ -529,6 +531,281 @@ func TestIsVersionValid(t *testing.T) {
...
+						},
+						{
+							Href: "../file.txt",
+							Rel:  "re-map",
+						},
+					},
+				},
+			},
+			curDir:        "/app",
+			newDir:        "/app/new",
```
mildwonkey: In these examples `newDir` is a subdirectory of `curDir` - does this method work the same way if the `newDir` is 
 entirely different - `/otherapp`, or if curDir is `/app/new` and `newDir` is `/app`?

src/pkg/common/oscal/component.go
```diff
@@ -102,6 +102,51 @@ func (c *ComponentDefinition) HandleExisting(path string) error {
...
 }
 
+// RewritePaths finds all the paths in the component definition relative to the baseDir and updates them to be relative to the newDir
+func (c *ComponentDefinition) RewritePaths(baseDir string, newDir string) error {
+	if c.Model == nil {
+		return fmt.Errorf("cannot remap paths, model is nil")
+	}
+
+	// Find all the paths in the component definition
+	pathsToRemap := []string{
```
mildwonkey: This is a very interesting way of handling this. It feels string-and-reflect heavy to me. I don't know if I want to nitpick this as much as use it as the starting point when designing a larger refactor; we already have types for all of these, I thought, so why use reflection instead of walking the c.Model?

mildwonkey (commented):
I feel a bit uncomfortable with this implementation, but not so much that I'd block it - it's partly based on not entirely understanding what's going on in here and partly based on my entirely fuzzy mental image of a future refactor, which isn't exactly a valid consideration 😂

meganwolf0:
> I feel a bit uncomfortable with this implementation, but not so much that I'd block it

Yeah I feel this too 😆 - I really wasn't sure _how_ to do this optimally/reusably...

So I think the bigger picture idea is that you can have an OSCAL model (this is just for component definition, but really any OSCAL Model) that can have links to other resources (remote or local). If we want to sort of package up all this data we may need to re-write the paths. Additionally, in the case of a component definition, it can import other component definitions (which contain these links) and those internal references will need to be re-written when the data gets imported to the importer model (ala, what we do with Compose right now, although those don't resolve paths, that function just puts all validations in the backmatter, which is not great for some of the use cases we've discussed with the SSP, etc).

The initial thought on this was to brute force look through all the spots where you'd possibly need to change a reference, and just do it. Although you can see from the list of items in `pathsToRemap` that it's a little on the long side. So totally agreed that reflect is not optimal, but that was my solution to not brute forcing it. Maybe there's another option that I'm overlooking if you have an idea on how to do it differently, I'm all ears!

src/pkg/common/common.go
meganwolf0: yeah this can be one line, this was a holdover for when that path was more complex but dumbed it down and left the old func - will fix!

mildwonkey:
My discomfort is because this feels like an implementation for iterating over raw json, but we have concrete types, so I'd expect something that iterated over each type.

Here's some probably-awful pseudo code:

func (c ComponentDefinition) RewritePaths (paths, anythingelse) error {
 for r := range c.Resources {
  if r.Href != "" {
    rewritePath(r.Href)
  }
  for s := range c.OtherField {
     // something like this for everything
  }
} 

Though for the actual implementation (and this is where my handwavey future ideas come into play, so again THIS MIGHT NOT MAKE SENSE for right now/ever) I'd probably have each type with links (every field inside the ComponentDefinition) fulfill an interface that includes a RewritePaths method, so we could handle each separately.

meganwolf0:
Ok fair enough, yeah agreed maybe it would be better to just write a super long function and have all the types checked by type and not with reflect. This does get a little tedious as well because of all the pointers... To be honest, this is probably something I could coerce chat-gipity to give me so I didn't have to write it manually 🤣 

For the future idea, I truly don't think I know enough about go-oscal to know if that's do-able. But maybe worth investigating more...

I suppose for the near term we're probably trading the two options: using the reflect implementation or write a big function. (selfishly/lazily I do think this implementation is easier to test/understand)

src/pkg/common/common_test.go
meganwolf0: It should work... I'll check this to see if there's a testing gap here

meganwolf0:
@brandtkeller if you have any weigh-ins on this TIA!

brandtkeller:
Reading back through the issue and then this implementation - I am weighing the "what problem does this solve now" with "are the surrounding assumptions correct". 

Yes the solutions complexity obfuscates some of the operations - I had to spend some time with debugging to get a clear-ish picture. 

But we have an original problem statement - composing component definitions that import other component definitions and remapping paths. Relevant problem for us - as each imported component-definition has potential links to Lula Validations or potentially other artifacts (which is why I appreciated the breadth of support in this PR). 

Apologies in advance. I believe we see the intent of a `package` operation for sake of transport - which allows us to dictate the artifact retrieval and replacement for re-mapping (evidence for a solution).

Do we see ourselves having any mechanism of getting away from a `compose` like operation? for component-definitions and allowing the import of other component-defintions - it feels like the answer is no and that this will be tied intimately to the execution of `validate` functionality - whether directly with Lula or with LaaL. Gut feeling here is that i'd default to a verbose solution that is easily traceable over a minimal solution that is less traceable.

meganwolf0:
@brandtkeller @mildwonkey - Updated this PR to remove the reflect method (as that sounded broadly unpopular 😅 ) in favor of the verbose model walk/rewrite in place

src/pkg/common/oscal/component.go
```diff
@@ -102,6 +102,279 @@ func (c *ComponentDefinition) HandleExisting(path string) error {
...
+// baseDir and newDir must be absolute paths
+func (c *ComponentDefinition) RewritePaths(baseDir string, newDir string) (err error) {
+	if c.Model == nil {
+		return fmt.Errorf("cannot remap paths, model is nil")
+	}
+
+	// BackMatter.Resources.Rlinks.Href
+	if c.Model.BackMatter != nil && c.Model.BackMatter.Resources != nil {
+		for _, resource := range *c.Model.BackMatter.Resources {
+			if resource.Rlinks != nil {
```
mildwonkey: I think this is a great change, it's much clearer for less-informed eyes ❤️. Explicit is better than implicit and clarity is kindness etc etc

I have some suggestions on how to make this a bit more readable. You could:
-  move the inner logic into smaller functions: `handleResourceLinks(c.Model.Backmatter.Resources)`
-  make functions that works on an array of links: `common.RemapPaths([]rlink.Href, baseDir, newDir)`
- make the functions methods on the type `Backmatter.RemapLinks(basedir, newDir)`

I think it's ok as it is for now, this isn't a blocker, just some ideas!

src/pkg/common/oscal/component_test.go
```diff
@@ -249,6 +249,7 @@ func TestMergeComponentDefinitions(t *testing.T) {
 		expectedImplementedRequirements       int
 		expectedTargetControlImplementations  int
 		expectedTargetImplementedRequirements int
+		uniqueComponent                       bool
```
mildwonkey: Is this only for tests, or is there some user-facing analogue to this?  Just for my own understanding

meganwolf0: oops this is what happens when I'm working on three branches at the same time, let me remove that and save for #893

brandtkeller (dismissed):
Thank you for the re-write - debugging traceability really improved my perspective. 

I'm a bit of a mess - this is all notional work not tied into execution currently - do we have issue(s) that add this to current execution?

brandtkeller:
> I'm a bit of a mess - this is all notional work not tied into execution currently - do we have issue(s) that add this to current execution?

#896 may be the execution implementation?

meganwolf0:
> #896 may be the execution implementation

That's right - trying to decide if that PR should also include instrumenting the method in the generate SSP function or if that should be separate. But that's really the endgame here: enable the SSP having the right links to pass through to the Assessment Plan (so then we can do a thing with that).

src/pkg/common/oscal/component.go
meganwolf0: Looking at these a little more, I think the last one is not going to be possible unless we change go-oscal - and I think the types are auto-generated so honestly I think doing like an interface there would be not super straightforward.

I think there are a few types here which are redundant across models (e.g., BackMatter, Metadata, ResponsibleRoles, Links) that could be compartmentalized into reusable functions...

mildwonkey (dismissed):

meganwolf0:
Alright, I know you all already approved this but I think @mildwonkey was right about refactoring that huge function (ultimately I do see reuse across models, so the goal was to make that more reusable/clear/testable) - with that, I decided to not be lazy and made the following updates:
* Broke down the ComponentDefinition `RewritePaths` method into smaller functions (that I put in `common.go` because they aren't all specific to the component definition model (I didn't break out every type, the intent was to break into types that I saw reused and/or ones that were nested to facilitate testing)
* Updated the test files for ^^ method - honestly it seemed unnecessarily confusing with the abs paths in there and hopefully the point still stands. I think import/validate will use real paths but mock paths I believe are sufficient to test this. After going through the above exercise, I also wanted to address gaps in testing all the elements of ComponentDefinition
* Added a fuzz to `RemapPath`, hopefully it's a valid implementation 😅

src/pkg/common/common_test.go
```diff
@@ -549,3 +656,11 @@ func FuzzReadValidationsFromYaml(f *testing.F) {
 		common.ReadValidationsFromYaml(a)
 	})
 }
+
+func FuzzRemapPath(f *testing.F) {
```
mildwonkey, line 669: 🎉 nice!!!

mildwonkey: 🤔 you may be able to round-trip this: (If this is valid) you could re-map back to the original path and add a check that last returned path is equal to the original test input. 

Something like this? 
gotPath := remap path (a, b, c) 
reversePath := (gotPath, c, b)

src/pkg/common/network/network.go
```diff
@@ -156,25 +156,45 @@ func FetchLocalFile(url *url.URL, config *fetchOpts) ([]byte, error) {
...
 		}
+	} else if url.Scheme != "" {
+		return ""
 	}
-	return ""
+
+	return handlePath(filepath.Join(url.Host, url.RequestURI()), baseDir)
+}
+
+func handlePath(path, baseDir string) string {
```
mildwonkey: This method name isn'g telling me anything, and it's not obvious to me why it returns an empty string if given an abs path but a full path to a directory if given a relative path. Can you add some comments that explains why this is the behavior? And maybe rename it -  `returnBaseDirIfNotAbsPath` seems like a more accurate description of what's happening (though I still wonder why)

src/pkg/common/oscal/component.go
```diff
@@ -102,6 +102,74 @@ func (c *ComponentDefinition) HandleExisting(path string) error {
 	return nil
 }
 
+// RewritePaths finds all the paths in the component definition relative to the baseDir and updates them to be relative to the newDir
+// baseDir and newDir must be absolute paths
+func (c *ComponentDefinition) RewritePaths(baseDir string, newDir string) (err error) {
```
mildwonkey, line 107: I love the changes here ❤️

mildwonkey (commented):
I love the changes! I left one incredibly minor comment on `handlePath` but I think this is looking great.

src/pkg/common/common_test.go
meganwolf0: Ok, I added this - it highlighted a thing where basically the path wasn't simplifying in the best way - e.g.,

If my starting path was ../app/file.txt, from the directory of /newapp, then you move it to /app, the file path still stayed the same (../app/file.txt) - which is _technically_ correct, however not really ideal. So I had to update the function. With this update, the test case "Invalid path remap" yields a different result... this is already an invalid path, but the way filepath.Join works it drops one of the "../", so the remap loses one as well... I'm kinda ok with this changing, since it was never a valid path, and since the purpose of this function isn't to actually stat the file, but just wanted to throw this out there that the change I made did affect tests and is probably a tradeoff of having a simplified path.

src/pkg/common/network/network.go
meganwolf0: Long story short on this one - there's an option to network.Fetch (`WithBaseDir`) that requires having an empty directory to work with non-relative files (absolute paths or remote URLs). This is quite possibly a bad design with the functions (perhaps something in there should return like a bool if relative or something, then remove the option), however that's not what's happening. To change this would trickle a bit more than I wanted to with this PR, and it goes into compose territory, so I just renamed and commented this function.

I'm hoping that's good enough for this and thinking this is something we can look at properly cleaning up when the import component defns modifications (see #810) get properly folded in.

mildwonkey (approved):
❤️ :shipit:

brandtkeller (approved):

PR #890, opened 2025-01-17
feat(api): support for non-json Content-Type

### LINKED ISSUE #883: [API] Support for non-json content types

Adding support for the API Domain to handle non-json content-types

- **Given** a validation with the API domain 
- **When** a validate is performed and the response returns `content-type` that is not json
- **Then** no error occurs and the data is made available as a string

### Expected Deliverable

- Modification to the API domain to support the conditional json unmarshall process.

brandtkeller:
## Description

Support for checking `Content-Type` - parsing json when applicable otherwise capturing `Raw` as a string value. 
## Related Issue

Fixes #883

## Type of change

## Checklist before merging


mildwonkey:
This looks good, I think you'll just need to update the e2e test fixture so it sends the correct content-type, and add one to capture the non-json content. 

dropping a random thought here, not a suggestion for this PR - There isn't an issue for this, since we didn't have any immediate need/use cases, but another potential future enhancement mentioned in the original design doc involves checking for content-encoding and decompressing responses first

mildwonkey (dismissed):
🎉 awesome!

lula-config.yaml
```diff
@@ -1 +1,8 @@
-log_level: info
\ No newline at end of file
+log_level: info
+
+constants:
```
meganwolf0: was this local test data/supposed to be pushed back to main?

src/pkg/domains/api/http_request.go
```diff
@@ -56,16 +57,24 @@ func doHTTPReq(ctx context.Context, client http.Client, method string, url url.U
...
 	if respObj.StatusCode >= http.StatusOK && respObj.StatusCode < http.StatusMultiStatus {
-		respObj.Raw = responseData
-		err = json.Unmarshal(responseData, &respObj.Response)
-		if err != nil {
-			message.Debugf("error unmarshalling response: %s", err)
-			return nil, err
+
+		// Check for the application/json response Content-Type
+		// Response is intended only for structured responses
+		if contentType == "application/json" {
```
meganwolf0, line 67: def not a blocking thing and probably could be easily updated in the future, since there are multiple content types that could resolve to json (e.g., `application/json; charset=utf-8`, `application/vnd.api+json`, etc) possibly worth thinking about supporting user defined values here

validation.yaml
```diff
@@ -0,0 +1,20 @@
+domain:
```
meganwolf0: similar question to config - is this file supposed to be here?

meganwolf0 (changes requested):

validation.yaml
brandtkeller: oops - that's embarrassing. Need to remove these

src/pkg/domains/api/http_request.go
brandtkeller: I agree - I'd like to clear the hurdle of allowing non-json responses in the most basic of forms but do think there is a broader question around expectations of a requests inputs and response outputs.

meganwolf0 (approved):

PR #894, opened 2025-01-22
feat: updated merge component defn

### LINKED ISSUE #893: Merge Component Definitions on UUID

Currently the `MergeComponentDefinitions` merges components based on unique titles, it would probably be better/more compliant with OSCAL for this function to be merging them on unique component UUIDs instead.

This function also need to be re-worked such that it can merge two component definitions where one does not have any components (which is valid oscal)

https://github.com/defenseunicorns/lula/blob/801551d3c34bd9f7c190c0a6de20e9b419d3c577/src/pkg/common/oscal/component.go#L121

### Additional context

None

### Expected Deliverable

- [ ] Updated function to use component.UUID as map keys + Updated tests (will need to re-write these)

meganwolf0:
## Description

Updates to the merge component definition function:
* Moves the merge of "oscal.DefinedComponents" out of the main function to allow merge on files with none defined (since this is valid OSCAL and is needed for the import component definitions methods)
* Changed merging on unique component titles to merging on unique component UUIDs

**Notes/Questions to Reviewers**
* Initially I thought the update to the merge on UUID would have to change using the `ComponentFromCatalog` method to create test data, but was able to bandaid that test method with a simple UUID copy if the component is not `unique`, as set in the test set-up. This is probably a bit flimsy, but works for our test cases... a potentially better solution would be to re-write that test to not use that method to generate data - I opted for the simpler approach but am open to revising it :)

## Related Issue

Fixes #893 

## Type of change

## Checklist before merging

- (n/a)[Schema Updates](https://github.com/defenseunicorns/lula/blob/main/docs/community-and-contribution/schema-updates.md) applied

### DIFF
--- a/src/pkg/common/oscal/component.go
+++ b/src/pkg/common/oscal/component.go
@@ -119,25 +119,50 @@ func MergeVariadicComponentDefinition(compDefs ...*oscalTypes.ComponentDefinitio
 
 // This function should perform a merge of two component-definitions where maintaining the original component-definition is the primary concern.
 func MergeComponentDefinitions(original *oscalTypes.ComponentDefinition, latest *oscalTypes.ComponentDefinition) (*oscalTypes.ComponentDefinition, error) {
-
-	originalMap := make(map[string]oscalTypes.DefinedComponent)
-
-	if original.Components == nil {
+	// Nil check on original and latest
+	if original == nil {
 		return original, fmt.Errorf("original component-definition is nil")
 	}
 
-	if latest.Components == nil {
+	if latest == nil {
 		return original, fmt.Errorf("latest component-definition is nil")
 	}
 
-	for _, component := range *original.Components {
-		originalMap[component.Title] = component
+	// merge the component-definition.components
+	if original.Components != nil && latest.Components != nil {
+		original.Components = mergeDefinedComponents(original.Components, latest.Components)
+	} else if original.Components == nil && latest.Components != nil {
+		original.Components = latest.Components
+	}
+
+	// merge the component-definition.back-matter resources
+	if original.BackMatter != nil && latest.BackMatter != nil {
+		original.BackMatter = &oscalTypes.BackMatter{
+			Resources: mergeResources(original.BackMatter.Resources, latest.BackMatter.Resources),
+		}
+	} else if original.BackMatter == nil && latest.BackMatter != nil {
+		original.BackMatter = latest.BackMatter
+	}
+
+	// Artifact will be modified - need to update the timestamp and UUID
+	original.Metadata.LastModified = time.Now()
+	original.UUID = uuid.NewUUID()
+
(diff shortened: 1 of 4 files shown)


src/pkg/common/oscal/component.go
```diff
@@ -119,25 +119,50 @@ func MergeVariadicComponentDefinition(compDefs ...*oscalTypes.ComponentDefinitio
 
 // This function should perform a merge of two component-definitions where maintaining the original component-definition is the primary concern.
 func MergeComponentDefinitions(original *oscalTypes.ComponentDefinition, latest *oscalTypes.ComponentDefinition) (*oscalTypes.ComponentDefinition, error) {
-
-	originalMap := make(map[string]oscalTypes.DefinedComponent)
-
-	if original.Components == nil {
```
mildwonkey, line 125: Since this method is now updating the `original` input pointer, you don't have to return the ComponentDefinition. While not strictly necessary, a side benefit is that it's clearer that we are modifying the original (so you don't use it later expecting it to _actually_ have the original).

For example, instead of writing

`new, err := MergeComponentDefinitions(original, latest)`
then doing stuff with `new`

we can just write
`err := MergeComponentDefinitions(original, latest)`
And now original is the merged version.

It's less about the clean writing and more about being very clear that `original` gets mutated by MergeComponentDefinitions, so we want to avoid future confused developers (me, probably) referring back to "original" thinking it's unmodified.

mildwonkey (commented):
I left some my usual (method signature, this time) nitpicks but the actual implementation all looks good!

src/pkg/common/oscal/component.go
meganwolf0: yeahhh so I didn't actually write this original function, agree with the comment on the signature. @brandtkeller - is this something we'd like to change? Happy to do so in this PR + all the associated places it gets used. The other thing I was considering was attaching this method to the ComponentDefinition itself and not having to pass. Would need to dig into what else that might affect, but I think that might be a better implementation?

meganwolf0: after fussing with this for entirely too long, I just ended up updating the function's signature, as requested. I was playing with adding as a method to ComponentDefinition, but that ended up messing with too much so rolled that back. Perhaps that's something we should do in the future though...

brandtkeller (approved):
No issues with current changes - I do want to call out that I believe we're missing some coverage around `capabilities` within the `component-definition` spec. Let's add an issue for that work as we haven't explored capabilities much.

meganwolf0:
> No issues with current changes - I do want to call out that I believe we're missing some coverage around `capabilities` within the `component-definition` spec. Let's add an issue for that work as we haven't explored capabilities much.

For posterity - just created sort of a broad issue to handle possible "capability" things - https://github.com/defenseunicorns/lula/issues/899

PR #896, opened 2025-01-22
feat: partial component compose

meganwolf0:
## Description

Adds method to `ComponentDefinition` which recursively imports data from `import-component-definitions` and remaps any links from those documents to be relative to the parent which imported said document.

**Notes to Reviewers**
* This method does not currently handle the templating aspect that is supported in `lula t compose`. I expect a follow-on PR to this to be to refactor that existing functionality to use this new method and then instrument the templating. Some additional thoughts might be that we extended the OSCALModel interface to more generically support templating, as that's probably something that could possibly be expected in other models? This might require further discussion/issue scoping.

## Related Issue

Relates to #810 

The fully closure of this issue, I expect, will be the implementation of this method in the `GenerateSystemSecurityPlan`, so that will probably be up next.

## Type of change

## Checklist before merging


src/pkg/common/oscal/component.go
```diff
@@ -170,6 +170,64 @@ func (c *ComponentDefinition) RewritePaths(baseDir string, newDir string) (err e
...
+// componentDir must be absolute paths
+// TODO: should componentDir be an attribute of the component definition?
+// TODO: Add templating
+func (c *ComponentDefinition) ImportComponentDefinitions(componentDir string) error {
+	if c.Model == nil {
+		return fmt.Errorf("cannot import component definitions, model is nil")
+	}
+
+	if c.Model.ImportComponentDefinitions == nil {
+		return nil
```
mildwonkey, line 184: What is this check for, and why isn't it an error? Given that you set this to nil at the end of the function, is it not valid to call it (inport comp defs from different directories) twice?

src/pkg/common/oscal/component.go
```diff
@@ -170,6 +170,64 @@ func (c *ComponentDefinition) RewritePaths(baseDir string, newDir string) (err e
...
+
+	// Add data from each to the current component definition
+	for _, importCompDef := range *c.Model.ImportComponentDefinitions {
+		// Create a new component definition from the imported component definition Href
+		importCompDefHrefAbs, err := filepath.Abs(filepath.Join(componentDir, importCompDef.Href))
+		if err != nil {
+			return err
+		}
+
+		data, err := os.ReadFile(importCompDefHrefAbs) // #nosec G304
```
mildwonkey: You can also validate the input and appease the linter with `filepath.Clean()` first - a good idea anyway, if it hasn't already been run on the link.

mildwonkey (commented):
I left some questions, but the ImportComponentDefinitions looks really great, I like how this all came together from the previous PRs!

src/pkg/common/oscal/component.go
meganwolf0: It's a parameter in the OSCAL Component Definition type (`import-component-definitions`) - sorry I realized after the fact that the method name also being called the same thing is confusing. Basically if the component is not importing any other component definitions we just do nothing here

src/pkg/common/oscal/component.go
meganwolf0: good call, I'll update that!

src/pkg/common/oscal/component.go
meganwolf0: I'm just going to update that method name :)

mildwonkey (approved):

brandtkeller (approved):
Looks great.

PR #897, opened 2025-01-23
chore(api domain): refactor spec processing

mildwonkey:
This PR separates the internal fields in the API Spec out into discrete types, moving the internal implementation details farther from the types that exist to unmarshal user input.

It looked weird and wrong (but does work) to have both the user-input Request and internal request types in the same package, so I moved the user types into schemas.

This work is part of #841

### DIFF
--- a/.github/workflows/scan-codeql.yaml
+++ b/.github/workflows/scan-codeql.yaml
@@ -42,7 +42,7 @@ jobs:
         uses: ./.github/actions/golang
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@f6091c0113d1dcf9b98e269ee48e8a7e51b7bdd4 # v3.28.5
+        uses: github/codeql-action/init@dd746615b3b9d728a6a37ca2045b68ca76d4841a # v3.28.8
         with:
           languages: ${{ matrix.language }}
           # config-file: ./.github/codeql.yaml #Uncomment once config file is needed.
@@ -52,7 +52,7 @@ jobs:
 
       - name: Perform CodeQL Analysis
         id: scan
-        uses: github/codeql-action/analyze@f6091c0113d1dcf9b98e269ee48e8a7e51b7bdd4 # v3.28.5
+        uses: github/codeql-action/analyze@dd746615b3b9d728a6a37ca2045b68ca76d4841a # v3.28.8
         with:
           category: "/language:${{matrix.language}}"
 
--- a/.github/workflows/scan-gosec.yaml
+++ b/.github/workflows/scan-gosec.yaml
@@ -42,6 +42,6 @@ jobs:
           retention-days: 5
 
       - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@f6091c0113d1dcf9b98e269ee48e8a7e51b7bdd4 # v3.28.5
+        uses: github/codeql-action/upload-sarif@dd746615b3b9d728a6a37ca2045b68ca76d4841a # v3.28.8
         with:
           sarif_file: results.sarif
--- a/.github/workflows/scan-kics.yaml
+++ b/.github/workflows/scan-kics.yaml
@@ -39,7 +39,7 @@ jobs:
           retention-days: 5
 
       - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@f6091c0113d1dcf9b98e269ee48e8a7e51b7bdd4 # v3.28.5
+        uses: github/codeql-action/upload-sarif@dd746615b3b9d728a6a37ca2045b68ca76d4841a # v3.28.8
         with:
           sarif_file: results.sarif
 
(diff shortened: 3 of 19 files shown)


meganwolf0:
Maybe just a few thoughts (because the PR lgtm, mostly just questions on the overall implications or precedence it sets)- 
* We have a few other domains and providers that follow a similar structure (storing their types in the domain/provider-specific location) - is a follow-on step to move those types to maintain consistency? I feel like it's weird if we just have the api spec types over in schema and everything else stays where it is
* We also have like the LulaValidation types defined, I suppose same question as above
* As somewhat discussed in slack, would another option be to move these definitions out of our library entirely, over to like /src/api/<v-whatever> to implement more separation/modularity across type changes over time
* Is this all OBE if we ever do a DSL?

I'm mostly of the opinion that we should do all or nothing, and not split up the definitions of the types (just to maintain internal consistency), but to move everything feels like a decent time investment.

mildwonkey:
To respond to your points in a rambling, out-of-order manner:

yeah, once (if) we do a DSL refactor, the layout will look very different. The parser might have the types like these, and at that point it all lives in a configs package and the domains are responsible for taking that input and mutating it. (maybe! there are other valid options!)

I wouldn't move the other domains' external types unless those domains were also refactored to have internal (non-exposed) structs, and the domain-specific `Spec` was only handled by a single logical path (here, it's the  CreateAPIDomain -> ValidateAndMutate -> output the "internal" types, never touch the Spec again). At some point I can see that all living in a configs (or, here's an idea, `domains/api/config`, if we want to move them back to the API but under a subpackage). The other domains use those  Spec structs directly in the domain, so I wouldn't move them before that change. (I'm a lot more comfortable with certain kinds of inconsistency, like this example, since it's backed by logic, but I do understanding not liking the inconsistency between domains. The implementation of all of them is sufficiently different though so I don't worry about it - maybe in a year or three we'll converge on a single pattern again, but pragmatically speaking I'd rather the code feel correct to me as a random go developer than as a lula developer who expects arbitrary patterns to be followed)

as to the LulaValidation types - I dunno 🤷🏻 ! I don't really know how and where they are used, what makes sense to you? I've lived in code bases where practically every package defined their own structs and converted (sometimes between nearly identical types) between packages, and terraform, which had a single "addrs" package that covered all resources over a huge swath of the codebase. 

So, uh ... wdyt? The only option I didn't particularly like was leaving the structs in the api domain, but I absolutely could just do that while we consider the options and see what the future brings.

meganwolf0:
Sorry, I'm back! Was trying to consider your response and see if others chimed in but yeah, this feels sort of like a compromise:

> or, here's an idea, `domains/api/config`, if we want to move them back to the API but under a subpackage

I agree on the types, and fair enough counter argument on the question of consistency. But seems like putting things under a config package, still within the particular domain, makes more sense than `schemas` (to me, anyway). Curious if @soltysh or @brandtkeller have a feeling either way.

brandtkeller:
I'm less opinionated about _where_, but do appreciate the intent on internal vs external types definition and the separation.

soltysh:
I'm personally leaning towards what @meganwolf0 expressed above, in slack and in that document of hers. Rather than doing ad-hoc moves just for these packages, I'd lean towards a major overhaul, with all the API bits landing in `src/api`, for example. Inside that we could have an internal* package with implementation details bits. I'd also encourage to separate function from the types, so inside `src/api/types.go` have only structures that could be consumed/used by lula users, and all the processing would be handled in separate packages. Basically `src/api` should be API only (although I will mentioned I'm biased from working with k8s APIs types :wink: [apps/v1 in k8s](https://github.com/kubernetes/api/blob/master/apps/v1/types.go)). 

* when introducing that internal package (whether you'll actually do [golang internal](https://go.dev/doc/go1.4#internalpackages) or not) make sure that outside consumers can fully operate on the exposed API. Ideally a test consuming that API and trying various operations, w/o the need for internal package.

mildwonkey:
I appreciate the feedback, but at this point I'm not sure what the best path forward is. We're moving towards a new architecture that isn't fully designed yet, which is why I wasn't _super_ worried about the impact of this decision in the long term - just about all of these types and packages will get shifted again anyway 🤷🏻
 
For now I'll keep all the types in api - I prefer to hold off on adding a config package until I'm actually working on the config refactor; I feel like I'm much more likely to get the package layout wrong (in ways that will annoy me in the long run) if I start moving things there now without the larger design in mind.

soltysh:
> For now I'll keep all the types in api - I prefer to hold off on adding a config package until I'm actually working on the config refactor; I feel like I'm much more likely to get the package layout wrong (in ways that will annoy me in the long run) if I start moving things there now without the larger design in mind.

That's roughly my thinking, I'm fully supportive of waiting for the re-arch work before tackling this refactor.

meganwolf0:
> waiting for the re-arch work before tackling this refactor

To this point, I'm not sure this is really fully accounted for in our roadmap this year, added a comment in that PR to hopefully address this so we can all be on the same page

brandtkeller (approved):
Appreciate the initiative and flexibility to feedback. I do like this separation and cleanliness.

